PR #15, opened 2014-07-29
Dev/algorithm config logging

linus-sherrill:


Purg:
So this has been open for a while. An incoming update massive update incorporates changes made here, rendering these modifications obsolete.

mleotta:
changes already accounted for in other merges

PR #54, opened 2015-05-05
Rix OCV plugin width/height return mixup

Purg:


doc/rel_notes/0.5.1.txt
```diff
@@ -27,3 +27,7 @@ Documentation
    added algorithms and containers and sorted the lists.
 
   * Added missing algorithm on the OCV module Doxygen main page and sorted.
+
+OCV Plugin
```
mleotta: [nit-pick] In v0.5.0, OCV is a module, not a plugin.

doc/rel_notes/0.5.1.txt
```diff
@@ -27,3 +27,7 @@ Documentation
...
+
+  * Flipped repective returns in width/height getter methods.
```
mleotta: [nit-pick] Might want to clarify that they were flipped because they were previously incorrect.

PR #64, opened 2015-06-29
Adjust test tolerances

mleotta:


Purg:
:+1: I don't have anything to say against this.

PR #65, opened 2015-06-29
update PPAs for Travis-CI

mleotta:
Travis-CI builds on Ubuntu 12.04, which has packaged versions of MAP-Tk dependencies which are too old for the latest MAP-Tk.  This branch adds PPAs which provide back ports of more recent versions of CMake, and Eigen.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ compiler:
   - clang
 
 before_install:
+  # CMake in Ubuntu 12.04 is too old. Get a newer version from this PPA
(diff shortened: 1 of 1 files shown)


Purg:
Is the test that's failing one of the ones that was addressed in the error-threshold-adjustment branch?

mleotta:
Sadly, no.  I have not seen that fail elsewhere, and it's only failing on the clang build on 12.04 (not gcc).  However, it very well could be another case of tolerances that are too tight, I will have to verify.

This branch does however get Travis much closer to passing on master.

PR #85, opened 2015-10-04
Dev/honor eigen3 root

kfieldho:
Allows Map-Tk to use the the EIGEN3_ROOT variable so that it more readily use Eigen as built by Fletch


kfieldho:
Looks like I based this on the wrong branch.

PR #86, opened 2015-10-04
Properly honor EIGEN3_ROOT if set

kfieldho:
This allows this Find module to work cleanly with super-builds
based on Fletch


mleotta:
I'm going to back-port this change to the release branch because it applied to older version of MAP-Tk as well.

kfieldho:
Do you mean master? If so, thanks, I was planning on submitting a separate
pull request. You've saved me a step.

Keith

PR #112, opened 2015-12-02
Switch to new Travis CI infrastructure

mleotta:
This patch changes the way dependencies are built for MAP-Tk.
We now use the new container-based infrastructure which is faster
but limits packages with can install with apt.  Instead, many packages
are built from source and then cached for future builds.

At this point the CI tests are only building a subset of the MAP-Tk components.


Purg:
:+1: Looks like it worked. Going to have to look into that caching section and scripts thing you have going here for SMQTK.

PR #141, opened 2016-03-25
fix pos2krtd

mleotta:


tools/pos2krtd.cxx
```diff
@@ -236,7 +236,7 @@ bool convert_pos2krtd_dir(const kwiver::vital::path_t& pos_dir,
   unsigned long num_files = dir.GetNumberOfFiles();
   for ( unsigned long i = 0; i < num_files; i++)
   {
-    kwiver::vital::path_t p = dir.GetFile( i );
+    kwiver::vital::path_t p = pos_dir + "/" + dir.GetFile( i );
```
Purg: Should we be using something to get platform-specific directory separator? Isn't there on in vital?

linus-sherrill: It appears that the windows API converts '/' to '\' in file paths. IIRC, Kwiversys code taken from CMake normalizes the paths by converting '\' to '/' for all systems.

Purg: Ok.

PR #144, opened 2016-03-28
default configs

mleotta:
Refactor sample configuration files into includable default algorithm configuration files that are installed.  Add top-level examples configurations files for a couple of public data sets.


Purg:
I think I lack the knowledge to weigh in on this.

mleotta:
@Purg You don't need to review this, unless you really want to.

PR #154, opened 2016-04-27
Make the link to ceres private.

dstoup:
Private linkage prevents the ceres dependency from leaking into projects that use Map-TK, but don't really need to care about linking ceres.

### DIFF
--- a/maptk/plugins/ceres/CMakeLists.txt
+++ b/maptk/plugins/ceres/CMakeLists.txt
@@ -32,7 +32,7 @@ kwiver_add_library( maptk_ceres
   )
 target_link_libraries( maptk_ceres
   PUBLIC               maptk
-                       ceres
+  PRIVATE              ceres
   )
 
 maptk_create_plugin( maptk_ceres


Purg:
I approve this message.

mleotta:
I merged this into master, but I think it should also go in release because it is more of a bug fix.  I will update release as well

mleotta:
@dstoup,  I think I merged this PR prematurely.  This has broken the dashboards on all platforms.  It turns out that Ceres should NOT be private in the maptk_ceres library because some of the public headers of this library include Ceres headers.  That said, most users of MAP-Tk should not be including these headers or directly linking to maptk_ceres or ANY of the libraries in plugins directory.  They should only link to main library and plugin loader.  The Ceres stuff get's loaded as a plugin at run time.  The only time you would use maptk_ceres directly is if you were creating your own plugin that builds on maptk_ceres (or if you were writing unit tests for the maptk_ceres internals, hence the dashboard failures)

Before I revert this change, maybe we should talk so I can better understand your issue.

dstoup:
I would say go ahead and revert and we can talk about the issues once the
dashboards are green. The issue is that packages that linked MapTK when
ceres was included had to also link ceres. We can maybe discuss better
options.

On Tue, May 3, 2016 at 9:57 PM, Matt Leotta <notifications@github.com>
wrote:

> @dstoup <https://github.com/dstoup>, I think I merged this PR
> prematurely. This has broken the dashboards on all platforms. It turns out
> that Ceres should NOT be private in the maptk_ceres library because some of
> the public headers of this library include Ceres headers. That said, most
> users of MAP-Tk should not be including these headers or directly linking
> to maptk_ceres or ANY of the libraries in plugins directory. They should
> only link to main library and plugin loader. The Ceres stuff get's loaded
> as a plugin at run time. The only time you would use maptk_ceres directly
> is if you were creating your own plugin that builds on maptk_ceres (or if
> you were writing unit tests for the maptk_ceres internals, hence the
> dashboard failures)
>
> Before I revert this change, maybe we should talk so I can better
> understand your issue.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/Kitware/maptk/pull/154#issuecomment-216719596>
>



-- 
David Stoup
R&D Engineer

Kitware, Inc.
28 Corporate Drive
Clifton Park, NY. 12065
518-881-4949 (W)
518-312-3946 (M)
518-371-4573 (F)

PR #161, opened 2016-05-20
Catch config_block exception and print it

laurennlam:
Catch config_block exception  from the main conf file when an included config file path is not found

### DIFF
--- a/gui/Project.cxx
+++ b/gui/Project.cxx
@@ -37,6 +37,7 @@
 #include <qtStlUtil.h>
 
 #include <QtGui/QApplication>
+#include <QtCore/QDebug>
 #include <QtCore/QDir>
 #include <QtCore/QFile>
 #include <QtCore/QFileInfo>
@@ -97,6 +98,12 @@ bool Project::read(QString const& path)
     }
 
     return true;
+  }
+  catch (kwiver::vital::config_block_exception e)
+  {
+    // TODO set error
+    qWarning() << e.what(); // TODO dialog?
+    return false;
   }
   catch (...)
   {


Purg:
I've had no part in the GUI stiff, but looks innocuous and fine to me. Maybe prefix the warning by mentioning the warning/error has to do with the configuration?

mleotta:
I agree with @Purg, it looks good to me, unless we want to take it further and make the message more verbose or possibly even add the dialog.  @mwoehlke-kitware, any thoughts before I merge this?

mwoehlke-kitware:
The intent, from what I can tell, is that `Project` should have an API to retrieve the error, when one occurs (see e.g. [QIODevice::errorString()](http://doc.qt.io/qt-5/qiodevice.html#errorString)), and the user (`MainWindow::loadProject`) would do something with it. This API doesn't exist yet, which is why there are a bunch of `TODO` comments.

For now, dropping the error message into `qWarning()` is fine, but I'd encourage keeping the `TODO set error` comment here also. `Project` itself should *not* show a dialog; that should be up to the user of `Project` how it wants to handle errors. (The `qWarning()` should also go away once `Project` has a retrievable error message.)

laurennlam:
Hi all,

I forgot to add an include to use the qWarning().
Can you wait until I correct my latest commit please ?

Laurenn LAM

Purg:
Theoretically, the continuous integration build using Travis-CI should hit that compilation issue and fail, not allowing us to merge anyway.

mleotta:
We will wait.  Travis-CI does not prevent a merge, I just lets you know when maybe it's not a good idea to merge because something is failing.

Purg:
Ah, I wasn't sure how it was configured for this project. :)

laurennlam:
The commit has been changed. Looks ok on my side.

Laurenn LAM

mwoehlke-kitware:
@laurennlam, did you see my comment re: `show dialog?` vs. `set error` for the `TODO`? (Otherwise LGTM...)

gui/Project.cxx
```diff
@@ -98,6 +99,11 @@ bool Project::read(QString const& path)
 
     return true;
   }
+  catch (kwiver::vital::config_block_exception e)
+  {
+    qWarning() << e.what(); // TODO dialog?
+    return false;
```
bastienjacquet: we should keep the `// TODO set error` for us to remember to handle the error when the API exists.

laurennlam:
@mwoehlke-kitware , I pushed the modification :)

PR #167, opened 2016-06-07
Dev/volume gui

ALouis38:
Added Depthmaps and volume visualisation to the GUI.

### Depthmaps visualisation

- Visualize depthmaps generated with PlaneSweepLib into World View and Depthmap View
- Load the depthmaps from a vti file list (generated by PlaneSweepLib). Added a parameter in the config file containing the path to the vti file list named depthmaps_images_file
- Filter the depthmaps based on the best cost values and uniqueness ratios
- Display the depthmaps as a point cloud or a surface in the World View
- Visualize the different arrays associated with the current depthmap in the Depthmap View

### Volume visualization:

- Visualize the volume generated with CudaDepthMapIntegration/Reconstruction into World View
- Load the volume from a vts file. Added a parameter in the config file containing the path to the vts file named volume_file
- Threshold the volume contour 
- Colorize the mesh using all or a subset of frames
- Display different coloring method


gui/vtkMaptkCamera.h
```diff
@@ -54,6 +54,15 @@ class vtkMaptkCamera : public vtkCamera
                     double (&projPoint)[2]);
 
   // Description:
+  // Project 3D point to 2D using the internal maptk camera
```
mleotta: need to update copyright year at the top of this file

mleotta:
@mwoehlke-kitware I could use some help in reviewing this MAP-Tk GUI branch from our friends at KEU.  I've been testing this branch out and I'm overall happy with the capabilities that it adds.  However, I could use your input on making sure it aligns with the design you had in mind.

Here is some initial feedback from me from a UI perspective.  Note, this is from testing on Mac OS X.  I'm not sure if the issues with not updating the display are specific to that OS.  We've run into similar issues in the past.

- I think the "Display Coordinates" options should be removed from the depth map menu and moved to an option under the "Show Grid" menu.  It should show measurements of the union of whatever the visible geometry is (landmarks, depthmap, volume, etc.)

- Depth map filters do not update until I force a redraw by rotating my viewpoint (which also causes the dialog to close).  It would be better to force a redraw of the display when "Apply" is pressed, or maybe even remove the "Apply" button and have it update on change (if not too slow).

- The "colorize surface" option under Volume does not apply when "current frame" is set until the frame is changed.  When "All frames" is set, the changing the "color display" mode does not change the display until the viewpoint is changed".  It would be nice to see it update right away.

- I'm not a big fan of adding a new "DepthMapView".  I would prefer that the depth map image visualization be just another overlay option in the Camera View.  However, We need to be able to show the depth map and image side-by-side, and can't do that without the ability to multiple camera views, each with their own view properties.  I'm will to accept the Depth Map View for now unless @mwoehlke-kitware thinks it would be simple to fold this into the Camera View and allow multiple Camera Views.

- I closed the depthmap view at one point, and now I don't know how to get it back.  Selecting it under the "view depth map" under the "View" menu does not make the pane come back like it does for the other removable/closable panes.

mleotta:
Thanks for addressing many of the UI issues from my last comment.  I'm testing again, and here are some more UI update-related issues I'm seeing:

1. When hitting the Apply button in the depthmap filters the 3D World View now updates, but the 2D DepthMap View does not update until the image is zoomed or panned.

2. When unchecking the depthmap filters checkbox, the filter settings continue to persist.  I would expect the filters to turn off when unchecking the box.  But the filters should not reset because you may want to toggle them back on with needing to change the values again.

3. This one is questionable: there may be value in the depth map filter settings persisting when the depthmap is changed.  That way I can set a threshold value and see the impact on the all the depth maps.

4. Checking the colorize surface checkbox does not actually enable colorization until the frame is changed or until the "current frame" radio button is clicked.

5. When unchecking the colorize surface option, I would expect the surface to return to the flat color shading that is used when it is first computed.

6. The "color display" option is reset when the compute button is pressed.  The colors always seems to be from "MeanColoration" after the compute button is pressed, but the combobox resets to "reconstruction_scalar".  I would expect the combo box to always display the color mode that is being rendered and that recomputing would not change the color mode.

What is "reconstruction_scalar"?  Is that just solid color?  If so, maybe you can take that out of the combobox and just have it use that mode when colorize surface is unchecked.

Thanks for all your hard work on these features.  This is really nice!

mleotta:
@ALouis38, it will help us review and merge this branch if you don't merge any new features into it.  For example, the webGL export is a separate pull request (which is great), but it currently has build issues and those build issues are propagated to this pull request.  It's best to keep them separate. You can still have an integration branch where you merge everything together for testing, but that should probably not be one of the branches that is used in a pull request.  It does make sense to merge the master branch back into this branch to resolve conflicts when they occur.

ALouis38:
This pull request has a lot of features integrated in it therefore I understand why it's difficult to review. I think it would be better if I close this one and cut it into smaller ones. It would be much more easy for everyone.

mleotta:
If you can factor out the changes from this branch into smaller branches I think that would be very helpful.  Though I realize that is potentially a lot of work.

mleotta:
Am I correct that the only feature remaining on this branch that is not in master now is the volume visualization?  Is that easy to factor out?

ALouis38:
There's also the volume and mesh exports, but we can count that in the volume visualization features. It shouldn't be that hard to factor out I think. Would you like me to close this pull request then?

mleotta:
Let’s wait until there is a pull request (or requests) for the remaining features.  Then we can close it.  I was just making sure I understood what was left.

mleotta:
I assume that PR #196 represents the remainder of this branch that has not been merged, so I'm closing this PR.

PR #170, opened 2016-06-16
"krtd_clean_up" option 

ALouis38:
Added a "krtd_clean_up" option in the bundle adjuster config file which can be turned to true in order to remove all the *.krtd files present in the "output_krtd_dir" directory before writing the bundle adjusted *.krd files. It's useful when the bundle adjuster is run several times and doesn't generate the same amount of *.krtd files each times.

### DIFF
--- a/tools/bundle_adjust_tracks.cxx
+++ b/tools/bundle_adjust_tracks.cxx
@@ -180,6 +180,9 @@ static kwiver::vital::config_block_sptr default_config()
                     "when updating cameras. This option is only relevent if a "
                     "value is give to the input_pos_files option.");
 
+
+      config->set_value("krtd_clean_up", "false",
+                        "Delete krtd files present into output_krtd_dir before launching the bundle adjustment.");
   kwiver::vital::algo::bundle_adjust::get_nested_algo_configuration("bundle_adjuster", config,
                                                      kwiver::vital::algo::bundle_adjust_sptr());
   kwiver::vital::algo::initialize_cameras_landmarks
@@ -1018,6 +1021,22 @@ static int maptk_main(int argc, char const* argv[])
   //
   // Write the output KRTD files
   //
+  if (config->get_value<bool>("krtd_clean_up") )
+  {
+    std::cout << "Cleaning " << config->get_value<std::string>("output_krtd_dir") << " before writing new files." << std::endl;
+
+    std::ifstream framelist(config->get_value<std::string>("image_list_file"));
+    std::string filePath, krtdFile;
+
+    while (framelist >> filePath)
+    {
+      krtdFile = ST::GetFilenameWithoutExtension(filePath) + ".krtd";
+
+      ST::RemoveFile(config->get_value<std::string>("output_krtd_dir") + krtdFile);
+    }
+
+  }
+
   if( config->has_value("output_krtd_dir") )
   {
     std::cerr << "Writing output KRTD files" << std::endl;
--- a/tools/bundle_adjust_tracks.cxx
+++ b/tools/bundle_adjust_tracks.cxx
@@ -41,8 +41,6 @@
 #include <string>
 #include <vector>
 
-#include <boost/filesystem.hpp>
-
 #include <vital/vital_foreach.h>
 #include <vital/config/config_block.h>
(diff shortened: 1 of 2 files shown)


tools/bundle_adjust_tracks.cxx
```diff
@@ -180,6 +180,9 @@ static kwiver::vital::config_block_sptr default_config()
                     "when updating cameras. This option is only relevent if a "
                     "value is give to the input_pos_files option.");
 
+
+      config->set_value("krtd_clean_up", "false",
+                        "Delete krtd files present into output_krtd_dir before launching the bundle adjustment.");
```
mleotta: For the description I would say

    Delete all previously existing KRTD files present in output_krtd_dir before writing new KRTD files

tools/bundle_adjust_tracks.cxx
```diff
@@ -1018,6 +1021,22 @@ static int maptk_main(int argc, char const* argv[])
   //
   // Write the output KRTD files
   //
+  if (config->get_value<bool>("krtd_clean_up") )
+  {
+    std::cout << "Cleaning " << config->get_value<std::string>("output_krtd_dir") << " before writing new files." << std::endl;
```
mleotta: Other status messages are going to std::cerr, this should probably be the same for consistency.

This is temporary.  In a future branch (pending release approval) these will all change to use the vital::logger anyway.

tools/bundle_adjust_tracks.cxx
```diff
@@ -1018,6 +1021,22 @@ static int maptk_main(int argc, char const* argv[])
...
+
+    std::ifstream framelist(config->get_value<std::string>("image_list_file"));
```
mleotta: This is not doing exactly what is promised by the description.  It is only deleting the KRTD files that have corresponding image files in the image list.  If the image list is changed you can still end up with extra KRTD files that are not deleted.  I think it would be better to just traverse all files in this directory and delete any files ending in ".krtd".

tools/bundle_adjust_tracks.cxx
```diff
@@ -1023,16 +1025,22 @@ static int maptk_main(int argc, char const* argv[])
...
+              << " before writing new files." << std::endl;
 
-    while (framelist >> filePath)
-    {
-      krtdFile = ST::GetFilenameWithoutExtension(filePath) + ".krtd";
+    boost::filesystem::path p(config->get_value<std::string>("output_krtd_dir"));
+    boost::filesystem::directory_iterator end_iter;
 
-      ST::RemoveFile(config->get_value<std::string>("output_krtd_dir") + krtdFile);
+    for( boost::filesystem::directory_iterator dir_iter(p);
```
mleotta: We don't want to pull Boost in for this.  We are trying to avoid needing to depend on Boost.  In this case it's easy.  There is already a function "files_in_dir()" implemented in this file around line 365.  You can just call that to get a list of files in a directory.

We are avoiding Boost by using [KWSys](https://github.com/commontk/KWSys).  There is a copy of KWSys in Vital named kwiversys.  You can do most of the same file path manipulations with that instead of Boost.

tools/bundle_adjust_tracks.cxx
```diff
@@ -1018,6 +1021,26 @@ static int maptk_main(int argc, char const* argv[])
...
+    std::cerr << "Cleaning "
+              << config->get_value<std::string>("output_krtd_dir")
+              << " before writing new files." << std::endl;
+
+    kwiver::vital::path_t krtd_dir = config->get_value<std::string>("output_krtd_dir");
+    std::vector<kwiver::vital::path_t> files = files_in_dir(krtd_dir);
+
+    for (int i = 0; i < files.size(); ++i)
+    {
+      if (ST::GetFilenameExtension(files[i]) == ".krtd")
```
mleotta: There is a subtle bug here that we've been bitten by before.  `ST::GetFilenameExtension()` returns the **longest** filename extension.  So if your file is named `image.0001.krtd` the extension returned is `.0001.krtd`.  Instead if you use `ST::GetFilenameLastExtension()` it would return `.krtd` for this same example.

Otherwise, this branch looks good.

PR #176, opened 2016-06-24
Launch Plane Sweep Algorithm from GUI

ALouis38:
A simple GUI for launching the plane sweep algorithm (depthmaps generation) of PlaneSweepLib directly from the Map-TK's GUI.
How to use:
* Load a project containing cameras and landmarks into mapgui.
* Go to Compute -> Depthmaps... -> Plane Sweep Lib
* Indicate the path to your version of PlaneSweepLib/build/bin/cudaPlanesweepMAPTk
* Compute

You can input any value for the different PSL parameters but the ones by default should work well.

### DIFF
--- a/gui/LaunchPlaneSweepView.cxx
+++ b/gui/LaunchPlaneSweepView.cxx
@@ -159,7 +159,7 @@ void LaunchPlaneSweepView::compute()
 {
   QTE_D();
 
-  std::string pslPath = d->UI.lineEditPSLPath->text().toStdString();
+  QString pslPath = d->UI.lineEditPSLPath->text();
 
   d->args << "--debug";
 
@@ -188,7 +188,7 @@ void LaunchPlaneSweepView::compute()
   {
     if (this->children().at(i)->inherits("QCheckBox"))
     {
-      QCheckBox *child = (QCheckBox*) this->children().at(i);
+      QCheckBox *child = qobject_cast<QCheckBox*>(this->children().at(i));
 
       if (child->isChecked())
       {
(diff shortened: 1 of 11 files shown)


gui/LaunchPlaneSweepView.cxx
```diff
@@ -0,0 +1,351 @@
+/*ckwg +29
+ * Copyright 2015 by Kitware, Inc.
```
mleotta: Copyright 2016

gui/LaunchPlaneSweepView.h
```diff
@@ -0,0 +1,77 @@
+/*ckwg +29
+ * Copyright 2015 by Kitware, Inc.
```
mleotta: Copyright 2016

gui/OutputDialog.cxx
```diff
@@ -0,0 +1,101 @@
+/*ckwg +29
+ * Copyright 2015 by Kitware, Inc.
```
mleotta: Copyright 2016

gui/OutputDialog.h
```diff
@@ -0,0 +1,64 @@
+/*ckwg +29
+ * Copyright 2015 by Kitware, Inc.
```
mleotta: Copyright 2016

mleotta:
I haven't reviewed the code, but I tested it and it works for me.  

This is a helpful tool to run things as they are now.  We can probably land this branch, but ultimately I'd like to see the Plane sweep command line tool go away.  Here is my high-level vision about where I'd like to see this go in the future.  An we will be helping with this at KUS starting soon.

 1. We write an abstract algorithm interface in KWIVER for depth map estimation that takes in a set of images with corresponding camera models, a depth range, etc. and produces a depth image, confidence (uniqueness) image, and best cost image.

 2. We modify the PlaneSweepLib clone that @bastienjacquet has to build a KWIVER plugin that implements this new API using the PlaneSweepLib code.  This plugin will be GPL and continue to live outside of KWIVER.

 3. We port the `cudaPlanesweepMAPTk` tool from PlaneSweepLib to MAP-Tk and revise it to use MAP-Tk and KWIVER data structures for images, cameras, etc.  This will call the new abstract algorithm for depth map estimation.

4. We modify this branch to directly call the depth map API from the GUI instead of an external executable.

5. We add an alternative implementation for depth estimation using Super3D that can now be called from either the command line tool or the GUI.

We'll get started with task 1 soon.

Your output dialog reminded me of something I've mentioned to @mwoehlke-kitware in the past.  I'd really like to see a dialog window like this that displays the log messages sent to the Vital logger so we can see the console output as the GUI is running algorithms.  This would be very helpful, especially on platforms like Windows.  I'm not suggesting that should be part of this branch, but maybe we can reuse this output dialog in the future.

PR #184, opened 2016-07-11
Integrate with kwiver

linus-sherrill:
Update map-tk code base to build with the new kwiver version that has all the algorithm plugins.


mleotta:
@linus-sherrill, can you remove the final commit on this branch (233134d) and force push?  We've already incorporated the white space fixes and the rest of the changes in that commit are erroneous.

mleotta:
actually, never mind.  I have another branch that does that and then adds more fixes on top.  I'll abandon this PR and make a new one.

PR #186, opened 2016-07-20
Undistortion

ALouis38:
Added a command line and GUI tools to undistort input frames. 
### Command line tool `maptk_undistort_frames`
The command line tool can be used in the same way as the others. It needs a configuration file (an example is provided in the config/ directory) with the list of the frames, the path to the directory containing the *.krtd files to use for the undistortion and the path to a directory for the undistorted frames. 
### GUI tool under `Compute > Undistort Frames`
The GUI tool take the cameras and frames loaded into the GUI as input. It only asks for a directory to save the undistorted frames.
### Vital fix
While working on this feature, I found out that when a *.krtd file was read, the distortion values weren't saved. This bug had to be fixed for those tools to work, and the fix is a pull request at Kitware/vital#25.
### Known issues

- The GUI output a trace of the undistortion progression in the console. The right thing to do here would be to output this in a dialog (something like the one in Kitware/maptk#176). I did not do this because I think it would be better to work on a generic output dialog later, possibly in another branch, so it could be reused.
- The example config file is isolated in an "other" directory. Maybe it would be better to have it integrated to one of the already existing examples, but I think it's not really a problem to have it in a separated directory.


mleotta:
From an architecture perspective, we don't want to make the maptk library depend on VTK.  The proper way to do this is to create an abstract algorithm in vital for image undistortion and then create a VTK plugin for MAP-Tk, that implements this API.  We need a VTK plugin anyway to as a place to eventually put the volume integration code and other VTK-based algorithms.  You should be aware that OpenCV also has functions to do this undistortion, and they have more options and are probably more efficient.  That's okay because there is nothing stopping us from having multiple back-ends to the abstract API.

The abstract algorithm in Vital should take a vital::image_container_sptr and a vital::camera_intrinsics_sptr and produce a warped image and updated camera_intrinsics.  The code that calls this function can deal with loading the appropriate images and cameras.  It should also write back out the updated cameras.

Note that there are many different ways to choose the updated intrinsics (called `perfectCam` in your code).  You don't always want to just clear the distortion coefficients.  That could cause part of the image to get warped outside the image bounds and be lost.  You may also need to adjust the image size, principal point, etc. It would be fine if the VTK implementation does the simple thing that you are doing and we add an OpenCV alternative later with more options.

For the bigger picture I would like to talk about the motivation for this branch with you and @bastienjacquet to make sure we have the same vision for how to handle image distortion in MAP-Tk. Undistorting the images is one path, but I'm also interested in having the algorithms and GUI work with the distortion model directly.  For example, in the GUI use the distortion coefficients when back projecting the image onto the ground plane and also when projecting 3D geometry into the camera view.

mleotta:
A warning should you decide to work on adding the Vital API and VTK plugin: there are big changes afoot to the MAP-Tk/Vital framework.  All of the MAP-Tk plugins are moving out of MAP-Tk and into the kwiver repository.  Vital is also moving into the kwiver repository.  This change is set to happen very soon, shortly after I make another release of MAP-Tk.  A preview of this change is here: https://github.com/kitware/kwiver/tree/next . The MAP-Tk plugins are called "arrows" in the new kwiver architecture.  After that next branch becomes master, we will land pull #184 which removes the plugins from MAP-Tk and requires MAP-Tk to build against the kwiver repo instead of Vital.

PR #196, opened 2016-08-02
Volume in world view

ALouis38:
A volume generated from the previous depth maps can now be loaded and observed in MAP-Tk.
The volume has to be a *.vts file, and the path to it has to be passed to the new parameter "volume_file".
### Features
- Hide/display the volume by clicking on the "Volume" button in World View.
- Modify the contour value to generate a mesh. By default (i.e. when the Volume button is first toggled), the contour value is 0.5.
- Enable different colorization modes for the generated mesh (current frame, all/a subset of frames, color by mean/median coloration, normals...)
- Export volume, mesh, or colored mesh in File > Export menu.


mleotta:
This seems to be missing `Helper.h`.  However, from what I can tell about what it contains, that code probably belongs elsewhere.  We don't want another function in the GUI to load KRTD files.  We should use the cameras that have already been loaded.

ALouis38:
> We don't want another function in the GUI to load KRTD files. We should use the cameras that have already been loaded.

Although it does include a function reading KRTD files, it has other function used by the reconstruction classes. 
Helper.h is an old file used back when the reconstruction wasn't part of MAP-Tk so it may still have some function already present in MAP-Tk. I'll edit that so we don't end up with duplicated functions.

gui/tools/Helper.h
```diff
@@ -0,0 +1,227 @@
...
+      pathList.push_back(std::string(globalPath) + "/" + elemsWithoutExtension[0] + ".krtd");
+    }
+
+    return pathList;
+  }
+
+  //----------------------------------------------------------------------------
+  // Description
+  // Read krtd file and create K and RT matrix
+  static bool ReadKrtdFile(std::string filename, vtkMatrix3x3* matrixK,
```
mleotta: I got a build warning that this function is unused.  That's good!  Let's remove it.

gui/tools/Helper.h
```diff
@@ -0,0 +1,227 @@
...
+      pathList.push_back(directoryPath + "/" + elems[elems.size() - 1]);
+    }
+
+    return pathList;
+  }
+
+  //----------------------------------------------------------------------------
+  // Description
+  // Read global path and extract all contained path
+  static std::vector<std::string> ExtractAllKRTDFilePath(const char* globalPath, const char* framelist)
```
mleotta: I got a build warning that this function is unused.  That's good!  Let's remove it.

gui/tools/Helper.h
```diff
@@ -0,0 +1,227 @@
...
+    else
+    {
+      return "";
+    }
+  }
+
+  //----------------------------------------------------------------------------
+  // Description
+  // Read global path and extract all contained path
+  static std::vector<std::string> ExtractAllFilePath(const char* globalPath)
```
mleotta: I got a build warning that this function is unused.  That's good!  Let's remove it.

gui/tools/Helper.h
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+
+  //----------------------------------------------------------------------------
+  // Description
+  // Compute median of a vector
+  template <typename T>
+  static void ComputeMedian(std::vector<T> vector, double& median)
+  {
+    std::sort(vector.begin(), vector.end());
```
mleotta: It's more efficient to compute the median using std::nth_element.  See for example 

http://stackoverflow.com/questions/1719070/what-is-the-right-approach-when-using-stl-container-for-median-calculation 

Technically the median requires averaging the two center elements for an even length vector.  However, in practice we often don't worry about that and just take closest index to the center.  e.g. `vector.size() / 2`

Ideally this helper function should move out of the GUI code, but that's not critical.

gui/tools/Helper.h
```diff
@@ -0,0 +1,227 @@
...
+    while (std::getline(ss, item, delim))
+      {
+      elems.push_back(item);
+      }
+  }
+
+  //----------------------------------------------------------------------------
+  // Description
+  // Get the filename path (copy from vtk in order to allow debug)
+  static std::string GetFilenamePath(const std::string& filename)
```
mleotta: If we need this still we should be using something from kwsys (in vital) or something from VTK or Qt.

gui/tools/Helper.h
```diff
@@ -0,0 +1,227 @@
...
+#include "vtkMatrix4x4.h"
+#include "vtkNew.h"
+#include "vtkTransform.h"
+
+namespace help
+{
+  //----------------------------------------------------------------------------
+  // Description
+  // Split string by separated char
+  static void SplitString(const std::string &s, char delim,
```
mleotta: It seems like we should have this in kwsys (vital), or elsewhere, but I haven't checked.

gui/tools/Helper.h
ALouis38: I've got this warning too, but it's used in ReconstructionData (ReconstructionData.cxx:74), which is used by MeshColoration, which is used by ColorizeSurfaceOptions. When I comment this function in Helper.h, I have build errors. 
Anyway, this function (and I assume the other ones you pointed at, even though I haven't checked yet) should disappear because they're duplicate of already existing ones.

mleotta: Yes, we certainly don't want two implementations of KRTD file parsing, so you should reuse the existing one in MAP-Tk.  However, it seems it would be even better if you didn't have to reload that cameras at all.  The cameras you need are the same ones that are already loaded.  It would be best to just access the existing data structures instead of reloading from the files, unless I'm misunderstanding something.

I do understand the historical reasons why this code is loading the cameras separately.

mleotta:
@ALouis38 @bastienjacquet I made a last minute decision to finally merge this branch to put into the MAP-Tk release I am making this week.  I addressed some of the comments I made previously by making the changes myself, but I did not address all issues.  I would still like to make this branch get cameras and images from the application rather than reloading them from files on disk, but some more work is needed to make this happen.   I decided to merge this anyway because I find myself needing this branch often to view final fused depth maps results.  I've been keeping this branch up to date on the side and merging it in when I needed it, but I'm tired of doing that.  We can continue to clean up and improve this functionality on new branches as we move forward.

bastienjacquet:
Sounds good to me. 
Let's continue to clean up and improve along the way, as time and missions permit :)

PR #199, opened 2016-08-10
attempt fixing travis build issue by adding blas/lapack install

Purg:


Purg:
Yeah, I changed it. When testing on a unbuntu 12.04 local docker instance, it wasn't being found. Had to specify it manually. Which now seems to show as working so far on travis.

Purg:
Now its tests that fail...

PR #201, opened 2016-08-16
Always build shared

mleotta:


linus-sherrill:
+1

CMakeLists.txt
```diff
@@ -45,12 +45,8 @@ include( maptk-depends )
...
 include(CMakeDependentOption)
 
-# we can't build shared libraries on Windows so we leave it off by default
-if (WIN32)
-  option(BUILD_SHARED_LIBS "Build shared libraries." ON)
-else()
-  option(BUILD_SHARED_LIBS "Build shared libraries." OFF)
-endif()
+# Always build shared libraries, required due to plugin infrastructure
+set(BUILD_SHARED_LIBS ON)
```
Purg: I think this still lets a user turn it off?  Might want to force setting into cache instead of / in addition to this.

mleotta: I copied this approach from Vital.  It seems to work.  I don't see the BUILD_SHARED_LIBS option in the CMake UI (ccmake) or in the CMakeCache.txt file.  Is that not what you would expect?

Purg: Just tested this out with vital. A user can specify `-DBUILD_SHARED_LIBS:BOOL=OFF`, making a cache entry that says OFF, but the build proceeds to build shared.  Part of me finds this confusing, and that a warning or something should be displayed when the user tries to build statically stating that we don't allow it.

PR #202, opened 2016-08-16
Find packages in Fletch

mleotta:


Purg:
May want to check that fletch was found or make it required when given a directory (not sure if that's done somewhere else).

May also want to list-append to CMAKE_MODULE_PATH in case there are system paths on it or something else added to it (https://cmake.org/cmake/help/v3.3/variable/CMAKE_PREFIX_PATH.html).

PR #221, opened 2017-02-01
Replaced matrix cast operator with .matrix() member function

mleotta:
This commit updates MAP-Tk with respect to change in the KWIVER API.
KWIVER replaced the Eigen matrix cast operator on some classes, like
vital::rotation, with a more explicit .matrix() member function.  This
was to avoid build errors introduced by API changes newer versions of
Eigen.


linus-sherrill:
Change looks good to me.

PR #256, opened 2017-08-04
[WIP] Upgrade to VTK 8.0

mleotta:
This branch should not be merged until VTK 8.0 is provided by Fletch.

As is, this branch compiles and run with VTK 8.0 and shows performance improvements when rendering large point clouds.


gui/CMakeLists.txt
```diff
@@ -165,18 +165,12 @@ target_link_libraries(TeleSculptor
   kwiver::kwiver_algo_core
   kwiver::vital_vpm
   vtkGUISupportQt
-  vtkInteractionStyle
+  vtkIOImage
   vtkIOPLY
   vtkIOXML
-  vtkViewsCore
```
mleotta: @RustyBlue I did not have an easy way of knowing which libraries were still needed, so I removed as many as I could such that it still successfully linked.  I assume several of these removed libraries are still being used and are linked as a result of dependencies of the VTK libraries I am still linking too.  Do you see any issue with removing these?  Should I leave some in to be more explicit?  Which ones?

mleotta:
@RustyBlue I've resolved the issue with the tracks not updating in the camera view when the camera frame is changed.  However, there is still an issue with a "double free or corruption" error when exiting the application that I need to investigate.

VTK 8.0 is now an option in Fletch `master` but is not the default version.

mleotta:
VTK 8 is now the default in Fletch

PR #277, opened 2018-04-12
Update MapTK with RPC camera changes

chetnieter:
This updates MapTK to make it compatible with changes to kwiver that will be introduced by https://github.com/Kitware/kwiver/pull/497. That pull request should be merged before this one.


chetnieter:
@mleotta Here are needed changes to MapTK to make it work with my initial RPC camera changes. This PR just switches vital::camera to vital::camera_perspective which means it will not be general to the base camera api. I think making things work with the base camera api should probably wait until after the upcoming public release.

PR #288, opened 2019-01-31
Handle track changes

mleotta:


gui/MainWindow.cxx
```diff
@@ -2670,6 +2677,31 @@ void MainWindow::updateToolResults()
     d->UI.actionKeyframesOnly->setEnabled(!d->tracks->tracks().empty());
     d->toolUpdateTracks = NULL;
   }
+  if (d->toolUpdateTrackChanges)
+  {
+    //SET THE TRACK FLAGS HERE
+    if (d->tracks)
+    {
+      for (auto const& change: d->toolUpdateTrackChanges->m_changes)
```
mleotta: @bclipp-kitware We should move this code to a helper function in KWIVER that applies track changes to tracks, but we can merge with it here for now.

PR #292, opened 2019-01-31
Update configs

mleotta:


config/ceres_bundle_adjuster.conf
```diff
@@ -66,10 +66,10 @@ ceres:loss_function_type = CAUCHY_LOSS
 ceres:max_num_iterations = 500
 
 # Number of threads to use in the linear solver
-ceres:num_linear_solver_threads = 2
+ceres:num_linear_solver_threads = 8
```
mleotta: @bclipp-kitware Do you recall if this actually made a performance difference?  At one point I tested this and the Ceres solver code could not effectively make use of more than 2 cores anyway.  However, that was on Linux and likely an older version of Ceres.

config/ceres_bundle_adjuster.conf
```diff
@@ -78,10 +78,10 @@ ceres:optimize_aspect_ratio = false
 ceres:optimize_dist_k1 = true
 
 # Include radial lens distortion parameter k2 in bundle adjustment.
-ceres:optimize_dist_k2 = false
+ceres:optimize_dist_k2 = true
```
mleotta: I don't think these settings change anything because above we have
```
ceres:lens_distortion_type = NO_DISTORTION
```
These "optimize" setting seem redundant, but the switch above tells you with model to use and these settings tell you whether to keep those parameters fixed in the optimization.  We should probably have k1 turned off as well for consistency.  It also doesn't do anything.

FYI, it's not a good idea to turn on distortion right now because the depth map estimation and depth map fusion do not use the distortion parameters (last I knew).  So if distortion is minor, it's better to have the best fit model without distortion so that dense operation are consistent.  This is something we need to fix before we can process really distorted data like GoPro.

config/gui_initialize.conf
```diff
@@ -3,5 +3,8 @@
 default_camera_config := gui_default_camera_intrinsics.conf
 
 block initializer
-  include core_initializer.conf
+  include keyframe_initializer.conf
 endblock
+
+initializer:core:interim_reproj_thresh = 2
```
mleotta: I guess this is not needed if it's also inside `core_initializer.conf`

config/keyframe_initializer.conf
```diff
@@ -0,0 +1,82 @@
...
+    
+    ceres:loss_function_type = SOFT_L_ONE_LOSS
+  endblock
+  
+  similarity_estimator:type = vxl
+
+  # Algorithm to use for 'camera_optimizer'.
+  # Must be one of the following options:
+  # 	- vxl
+  camera_optimizer:type = vxl
```
mleotta: There is a Ceres camera optimizer as well that should be more efficient.

config/keyframe_initializer.conf
```diff
@@ -0,0 +1,82 @@
...
+
+  # Algorithm to use for 'essential_mat_estimator'.
+  # Must be one of the following options:
+  # 	- vxl
+  essential_mat_estimator:type = vxl
+  
+  estimate_pnp:type = ocv
+
+  # The number of samples to use in RANSAC
+  essential_mat_estimator:vxl:num_ransac_samples = 500
```
mleotta: @bclipp-kitware You dropped this to 100 in the core initializer algorithm.  I recall 100 not be enough for a robust solution but it was painfully slow at 500.  Should I change the other one back to 500?  Should I change this one to 100?  Is there a reason they should be different?

config/keyframe_initializer.conf
```diff
@@ -0,0 +1,82 @@
...
+  lm_triangulator:core:homogeneous = false
+  lm_triangulator:core:min_angle_deg = 1
+
+  # Algorithm to use for 'lm_triangulator'.
+  # Must be one of the following options:
+  # 	- core
+  # 	- vxl
+  lm_triangulator:type = core
+  
+  block global_sfm
```
mleotta: @bclipp-kitware This is not being used right?

PR #293, opened 2019-02-04
The type 'uint' is not defined by all compilers, use 'unsigned int'

mleotta:


mleotta:
already done in another branch

PR #307, opened 2019-06-03
Fixes loading of ground control points on MacOS.

chetnieter:
Removed range transform adapter as it was not correctly parsing the json data from the GCP file.


mleotta:
Looks good to me.  I will merge after CI succeeds.

PR #321, opened 2019-06-07
Minor fix to path scripts

chetnieter:
Minor fix to path scripts so version number in paths get set
correctly.


chetnieter:
@mleotta PTAL - I noticed the path scripts were not getting the version number in some of paths.

mleotta:
Thanks!

PR #328, opened 2019-06-12
Add Dockerfile

mleotta:


mleotta:
@chetnieter Here is the Dockerfile for building the Ubuntu package.  It seems to work now.

PR #351, opened 2019-11-27
Logger window

mleotta:
This branch requires Kitware/kwiver#904

### DIFF
--- a/doc/release-notes/master.txt
+++ b/doc/release-notes/master.txt
@@ -12,6 +12,8 @@ TeleSculptor Application
 
  * Updated the code to work with API changes to kwiver::vital::geo_point.
 
+ * Added a logger panel to show log message.
+
 
 Fixes since v1.0.0
 ------------------
--- a/gui/CMakeLists.txt
+++ b/gui/CMakeLists.txt
@@ -82,6 +82,7 @@ set(gui_ui
   DepthMapViewOptions.ui
   FeatureOptions.ui
   ImageOptions.ui
+  LoggerView.ui
   PointOptions.ui
   VolumeOptions.ui
 )
@@ -111,6 +112,7 @@ set(gui_sources
   GroundControlPointsWidget.cxx
   GuiCommon.cxx
   ImageOptions.cxx
+  LoggerView.cxx
   MainWindow.cxx
   MatchMatrixAlgorithms.cxx
   MatchMatrixWindow.cxx
--- a/gui/LoggerView.cxx
+++ b/gui/LoggerView.cxx
@@ -75,9 +75,10 @@ void LoggerView::logHandler(kv::kwiver_logger::log_level_t level,
                             kv::logger_ns::location_info const& loc)
 {
   QTE_D();
-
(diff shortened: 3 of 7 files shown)


mwoehlke-kitware:
BTW, have you tested how well this performs with a few thousand messages? (Or is that unlikely to happen in practice?)

mleotta:
We have multiple threads throwing messages at this. There are thousands of
messages but not all simultaneously.  Maybe tens of messages per second.
Overall it doesn't feel slow, but I haven't timed it.

mwoehlke-kitware:
I was wondering, more, once the log is thousands of messages long, does it still scroll without bogging down? I know KDevelop uses something different for logging, and can't help wondering if there is a reason they don't just use a QPlainTextEdit...

mleotta:
Good question.  I'll have to do some more testing.  I haven't really let
the log fill up yet.

FYI, at some point in the future I'd like to add filters in here to show
only logs at a certain level or from a certain source.  That will change
how this works.  For a first pass I'm happy to just see everything.  It's
better than seeing nothing, which is what we have without this branch.

mwoehlke-kitware:
Hmm... if you (know you will) want filtering, I would *strongly* consider investing the effort up front to create a [QAbstractListModel](https://doc.qt.io/qt-5/qabstractlistmodel.html) rather than using QPlainTextEdit. The disadvantage is you lose trivial copy/paste and partial selection, but I suspect QListView will perform better since it has built-in line optimization, especially if you aren't wrapping text and can leverage uniform row height. This is also almost surely an instance when you definitely want to use QList rather than QVector as your backing store.

mleotta:
I was wondering whether I should put this data into a table-like structure for display.  The reason I haven't gone down that path yet is that I do sometimes have long, multi-line log messages including some from Ceres that write out status data in somewhat structured text (i.e. numbers aligned in columns).  So I think this would be difficult to read if formatted in a table.  Currently I'm using plain text for display with a monospace font to keep things aligned.  I've started to add HTML tags to visually highlight the source of the log message.  It seems to be working well.

In the future I may cache all log messages in list or vector and if filters are selected, only pass through a subset to the QPlainTextEdit display.  If the filters are changed I will clear the text box and refill it with data from the cache while applying the new filter.  That may be slow, but changing filter setting probably will not happen often.  It should only be a problem if it becomes painfully slow.

mleotta:
@mwoehlke-kitware can you help me with the error in the travis builds? I see this warning:

    qte-amc: /home/travis/build/Kitware/TeleSculptor/gui/LoggerView.ui: warning: no actions found; no output will be produced

I get the same warning on Windows, but it still builds.  On linux it seems to not make the other "am_*.h" files when this warning happens so CameraView.h can not find "am_CameraView.h".

What is the right way to add the LoggerView in CMake for QtExtensions if there are no actions (yet)?  Or do I need to add some sort of action to the LoggerView.ui?

mwoehlke-kitware:
If there are no actions yet, just don't AMC-wrap it.

PR #355, opened 2020-03-03
Track features

danlipsa:
It needs:
https://github.com/Kitware/kwiver/pull/948


danlipsa:
@mleotta @linus-sherrill Can I use a kwiver branch here to make this pass or I have to wait until kwiver is merged?

linus-sherrill:
Jenkins, test this please

danlipsa:
@mleotta I added back maptk_track_features

danlipsa:
I updated kwiver to the latest and updated the copyrights. Ready to merge?

mleotta:
@danlipsa one problem we will have with this branch is that the GUI and tools will no longer be able to find the config files that have moved to KWIVER.  Maybe for now we should leave a copy of these config files in TeleSculptor to get this branch merged, and then address the config file relocation in a new branch.

When packaging the TeleSculptor in `gui/CMakeLists.txt` we will need to specifically install the config files from KWIVER.  We will then need to update the code to find those config files.  This is non-trivial because the current helper function is designed to look only in paths for the application (TeleSculptor in this case).  See for example:
https://github.com/Kitware/TeleSculptor/blob/master/gui/Project.cxx#L106

The installed application does not normally use the `$KWIVER_CONFIG_PATH` environment variable.  Instead it relies on a pre-configured search path based on the application name and version.  We will need to extend this to search first for TeleSculptor configuration files and then for KWIVER configuration files.  We probably need to modify [read_config_file()](https://github.com/Kitware/kwiver/blob/master/vital/config/config_block_io.cxx#L240) in KWIVER
so that it calls `application_config_file_paths()` twice.  Once for the application name/version and once for "kwiver" and its version.  We could then concatenate these search paths.

danlipsa:
@mleotta Thanks for the pointers. I'll copy the configs back in TeleSculptor to merge and make the changes as you outlined in the subsequent MR.

danlipsa:
@mleotta ping

PR #361, opened 2020-05-07
Move Configuration files into KWIVER

mleotta:


mleotta:
@danlipsa This branch picks up what you started with moving configuration files to KWIVER.  The config files have been removed (again) and KWIVER has been updated such that TeleSculptor is able to find config files in KWIVER.

PR #363, opened 2020-05-19
Revert camera line width

mleotta:
This fixes the bug in which cameras were not draw on Linux


mleotta:
@danlipsa see if this fixes the camera visibility issues you were seeing

PR #366, opened 2020-06-01
Speed up playback

mleotta:
Playing back video in TeleSculptor was slower than necessary because advancing the video is always done by sending a message to go to a specific frame number. Previously this meant that seek was called for each frame change, even if the frame number was just stepping to the next frame.  Seek calls step the video back to the previous key frame and then forward again to reach the target frame.  It is more efficient to first check if the video is already just a few frames behind the target and then advance the video in that case instead of seeking.

### DIFF
--- a/gui/MainWindow.cxx
+++ b/gui/MainWindow.cxx
@@ -1144,10 +1144,18 @@ void MainWindowPrivate::loadEmptyImage(vtkMaptkCamera* camera)
 //-----------------------------------------------------------------------------
 void MainWindowPrivate::loadImage(FrameData frame)
 {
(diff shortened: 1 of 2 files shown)


gui/MainWindow.cxx
```diff
@@ -1147,7 +1147,16 @@ void MainWindowPrivate::loadImage(FrameData frame)
   // TODO: check if seek vs next_frame is needed
```
chetnieter, line 1147: This TODO can be removed.

mleotta: done

PR #369, opened 2020-06-02
Make metadata scan optional

mleotta:
This is the last feature change I would like to make before releasing TeleSculptor v1.1.0.  This is a big improvement to usability.  Previously we required a scan of the video to count frames and load metadata each time it was opened.  This could take a full minute, or in some cases much longer.  With a recent KWIVER change we can now get an accurate frame count in constant time.  So we can now immediately load everything except the metadata map in a few seconds.

The `VideoImporter` still does a scan to build the metadata map, but this is now cancel-able.  The algorithms can still run without the metadata if you cancel, or you can wait and run with metadata.

### DIFF
--- a/doc/release-notes/master.txt
+++ b/doc/release-notes/master.txt
@@ -2,7 +2,7 @@ TeleSculptor v1.1.0 Release Notes
 =================================
 
 This is a minor release of TeleSculptor that provides both new functionality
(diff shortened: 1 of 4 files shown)


mleotta:
@chetnieter I know you are busy, but if you could find time to take a quick look at this branch it would help.  This is a modification to the VideoImporter that you wrote.

PR #374, opened 2020-06-15
Documentation updates

russelldj:
I tried to do this as somewhat distinct updates so you could cherry pick them. I don't think I corrected anything which was actually incorrect but I hope these changes would make it easier for a novice user to follow the project. I'll add a few more updates tomorrow.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -78,25 +78,6 @@ project from source or run command line tools.  For end users looking for
 instruction on running the GUI application please read the
 `User Guide <doc/TeleSculptor-v1.0-User-Guide.pdf>`_.
 
-Overview of Directories
-=======================
-
-======================= ========================================================
-``CMake``               contains CMake helper scripts
-``config``              contains reusable default algorithm configuration files
-``doc``                 contains release notes, manuals, and other documentation
-``examples``            contains example tool configuration for public datasets
-``gui``                 contains the visualization GUI source code and headers
-``gui/icons``           contains the visualization GUI icon resources
-``maptk``               contains the maptk library source and headers
-``packaging``           contains support files for CPack packaging
-``scripts``             contains Python helper scripts
-``plugins/blender``     contains Python plug-ins for Blender
-``plugins/sketchup``    contains Ruby plug-ins for SketchUp
-``tests``               contains testing framework and tests for each module
-``tools``               contains source for command line utilities
-======================= ========================================================
-
 Installation
 ============
 If you have downloaded an installer from the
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -94,37 +109,66 @@ minimum required version of CMake is 3.9.5, but newer versions are recommended.
...
 
-  # On Linux systems, Install the following packages before building
-  $ sudo apt-get install build-essential libgl1-mesa-dev libxt-dev
-  $ sudo apt-get install libexpat1-dev libgtk2.0-dev liblapack-dev
+  sudo apt-get install build-essential libgl1-mesa-dev libxt-dev
+  sudo apt-get install libexpat1-dev libgtk2.0-dev liblapack-dev
+
+
+Set up the folder structure and obtain the source files. This can be done with
+git or by downloading the files  and extracting them. Then setup the folder(s)
```
mleotta, line 129: nitpick: double space between "files" and "and".

README.rst
```diff
@@ -94,37 +109,66 @@ minimum required version of CMake is 3.9.5, but newer versions are recommended.
...
+  # configuration options
+
+Generate the makefile/msvc solution to perform the superbuild using cmake.
+A description of the configuration options can be found in `CMake Options`_.
 
-  # Run CMake
+.. code-block :: bash
+
+  # From the build directory provide cmake the path to the source directory,
+  # which can be relative or relative or absolute.
```
mleotta, line 157: relative mentioned twice.

README.rst
```diff
@@ -137,6 +181,27 @@ Here is the quickest way to build via a cmd/bash shell
   # To edit TeleSculptor code, open the
   # build/external/telesculptor-build/TeleSculptor.sln
 
+This will produce an installer file which can install TeleSculptor in the same
+way as the downloadable installer, as detailed in the next section.
+
+Installation
```
mleotta, line 187: I'm wondering if this section should go after all the build options (before MAP-Tk Tools).  As it is, we talk about installation and then go back to talking about build options, which seems odd.  

Maybe an even better option is to put the installation section before the `Build TeleSculptor` section.  Since I expect most users would just download and install this.

README.rst
```diff
@@ -137,6 +181,27 @@ Here is the quickest way to build via a cmd/bash shell
...
+------------
+Once you have an installer downloaded or built, you need to run it to install
```
mleotta, line 189: These directions are specific to the Linux self extracting installer.  To cover all bases we should talk about the others

On Windows run the installer executable (exe) and follow the prompts in the installer dialog.  Administrative permission is required.  On Mac, open the disk image (dmg), accept the license terms, then drag the TeleSculptor application into the Applications folder.  On Linux, run the self extracting installer script (sh) ...

mleotta:
One other thing.  We should target branches to merge into `release` that are intended to land in the upcoming release.  I've already changed the target for this one.

README.rst
russelldj: I'll include this. Do we additionally need to update the section `Before building on Linux systems...` to include instructions for installing dependencies on other OS or are they not required?

README.rst
russelldj: I agree that it's currently in a weird place and think it could go before build. However, in that case, I don't think `Overview of Directories` makes sense because I don't think many end users will care about that. I'm messing around with other places to put it now, in addition to some formatting issue from my incompetence at reStructuredText.

README.rst
mleotta: We should try to enumerate dependencies on other platforms at some point and describe how to install them, but it's not critical right now.  Generally the dependencies are fewer on other platforms.  You need your build tools (e.g. Visual Studio) and you probably need to install Python with docutils if you want the documentation.  However, there is no single standard way to install these things on Windows and Mac.

russelldj: I think multi-platform installation is addressed better now.

README.rst
russelldj: It's moved. I was pretty heavy handed in pointing build-from-source users to the right sections so I don't think there should be any confusion there.

russelldj: As a separate commmit I moved `Overview of Directories`. I don't think that's critical but I do think it reads better like that.

russelldj:
I don't see any more changes I'd like to make before the release unless you would like me to remove the mentions of examples. I don't think that I can effectively update examples especially without being able to run TeleSculptor.

PR #376, opened 2020-06-23
Added new TeleSculptor README

russelldj:
Updated the example README. Would still like to reorganize and eventually add images.


russelldj:
Something is broken in the build, I assume related to removing the examples directory.

russelldj:
It builds successfully with the latest commit.

CMakeLists.txt
```diff
@@ -106,9 +106,6 @@ add_subdirectory(config)
 # this is where the plugins to 3rd party software live
 add_subdirectory(plugins)
 
-# this is where configuration for example data sets lives
-add_subdirectory(examples)
```
mleotta, line 110: I think we should leave this in along with the CMakeLists.txt file in the examples directory.  This just installs whatever is in the examples to include it with the installer.  For now that would just be the EXAMPLES.rst, but that's okay.  We may add more later.

README.rst
```diff
@@ -303,7 +303,6 @@ Overview of Directories
 ``CMake``               contains CMake helper scripts
 ``config``              contains reusable default algorithm configuration files
 ``doc``                 contains release notes, manuals, and other documentation
-``examples``            contains example tool configuration for public datasets
```
mleotta, line 306: maybe we leave this, but say "contains pointers to example public datasets to use"

doc/EXAMPLES.rst
```diff
@@ -0,0 +1,69 @@
...
+Additionally, the way to data is captured has a significant impact on the
+quality of the model. It is important to have multiple views of the same regions
+from different directions and if possible the camera should complete a full rotation about the scene.
+It is best not to change the zoom of your lens and if you do it should be done
+slowly and gradually. Finally, you should avoid switching between different
+modalities, for example recording part of your video in color and the rest in infrared.
+If you are capturing images rather than video, it is important that there is
+substantial overlap in the content seen between different images.
+
+Setup
```
mleotta, line 39: These next three sections are redundant with user guide.  This is all good stuff, but I'm sure we will forget to maintain this text as TeleSculptor is updated.  It's probably better to have this information in a common spot, which is the user guide you've already linked to above.  I would remove the sections below.

In the future we should probably make the user guide something other than a PDF (generated from a Word Doc), but that's a topic for another day.

examples/CMakeLists.txt
```diff
@@ -1,7 +0,0 @@
-# Install config files
```
mleotta, line 1: undelete this file as described above

doc/EXAMPLES.rst
russelldj: I should have read the user guide first. Could the `Data` section just become part of the README since it is small and isn't associated with any other files?

mleotta: Yes, I suppose it could. As written it is fairly brief and doesn't provide additional files.  However, I think over time I'd like to add more files here.  This would include a still image from each video and maybe a resulting mesh file.  We may add other video sources later too.  So maybe it's best just to keep it in it's own folder for now.

russelldj:
I believe I fixed everything you mentioned.

PR #419, opened 2021-02-16
Support VTK 9.0

mleotta:


mleotta:
@borovik135 you might need this to build with VTK 9

mleotta:
@sankhesh does this look like an appropriate way to support both VTK 8.2 and VTK 9 in TeleSculptor?

sankhesh:
@mleotta The changes look good to me.

You need to add the `vtk_module_autoinit` macro for VTK 9 in CMakeLists.txt. See https://kitware.github.io/vtk-examples/site/Cxx/Plotting/StackedPlot/#cmakeliststxt for an example.

mleotta:
Thanks.  What does `vtk_module_autoinit` do?  It seems to be working without it.  Is that only needed if I link to `${VTK_LIBRARIES}` or libraries using the `vtkName` pattern rather than `VTK::Name`?

sankhesh:
> What does vtk_module_autoinit do?

It is a way to manage object factory registration between different modules. See the details here: https://vtk.org/doc/nightly/html/md__home_kitware_dashboards_buildbot_vtk_nightly-master-ike-linux-shared-release_doc_nightly_osm749dd663df9981384f2598c108aac3b0.html

PR #428, opened 2021-03-12
Add Flatpak

kevinsmia1939:
This Flatpak manifest still does not build correctly. But hopefully will get more attention in the future.


mleotta:
If we could download at build time that would make this whole process a lot easier and easier to maintain as well.  Otherwise as package versions change in fletch we will need to update them in the flatpak manifest as well.  It will be easy for these to get out of sync.  I came across this post about downloading at build time in flatpak:

https://casept.github.io/post/flatpak-download-at-buildtime/

Would this work here?

kevinsmia1939:
> me that would make this whole process a lot easier and easier to maintain as well. Otherwise as package versions change in fletch we will need to update them in the flatpak manifest as well

It could work, but I think it might now be allow to submit to Flathub, they really care about reproducible build.

PR #441, opened 2021-04-06
User Guide converted to RST

jacobderosa:
The user guide was converted from a PDF to RST such that it can be viewed on readthedocs.


mleotta:
Can you access this link?
https://readthedocs.org/projects/telesculptor/builds/13441438/

or maybe this one: https://readthedocs.org/api/v2/build/13441438.txt

The build failed.

mleotta:
Oh, it links it right from the Github PR as a CI test, very cool.  So you
should be able to iterate on the PR until it works.

mleotta:
I don't think we should check in the generated files in `_build`

jacobderosa:
The build failed because it didn't know to look for index.rst. It was looking for contents.rst, which is the default, but it should be fixed now. I'll remove the _build folder.

doc/conf.py
```diff
@@ -0,0 +1,55 @@
...
+# This pattern also affects html_static_path and html_extra_path.
+exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
+
+
+# -- Options for HTML output -------------------------------------------------
+
+# The theme to use for HTML and HTML Help pages.  See the documentation for
+# a list of builtin themes.
+#
+html_theme = 'alabaster'
```
mleotta, line 47: Can you change the theme to "sphinx_rtd_theme"?

doc/manuals/advancedconfig.rst
```diff
@@ -0,0 +1,97 @@
...
+
+Note that the "max_frames" parameter is in the "feature_tracker" scope and scope must be specified either using block/endblock notation or with a prefix before a colon.
+
+Printing all KLV metadata
+===========================
+
+The TeleSculptor application loads KLV metadata and display it in a viewer, but there is no way to export this data in batch.  However, the installer does provide the kwiver command 
+line tool that has an applet that will print out all metadata in a video.  This applet is called “dump_klv”. The default installation path is
+
+**C:\Program Files\TeleSculptor 1.1.0\bin\kwiver.exe**
```
mleotta, line 89: The `\` character is an escape character and so these paths don't format correctly.  I think you need to use one of the following:

    **C:\\Program Files\\TeleSculptor 1.1.0\\bin\\kwiver.exe**
    ``C:\Program Files\TeleSculptor 1.1.0\bin\kwiver.exe``

or in a literal block like this

    The default installation path is::

        C:\Program Files\TeleSculptor 1.1.0\bin\kwiver.exe

note the `::` and the indentation.

mleotta: I've been thinking that we might also want to replace all instances of the current TeleSculptor version with a substitution like this

    **C:\\Program Files\\TeleSculptor |version|\\bin\\kwiver.exe**

    **“C:\\Program Files\\TeleSculptor |version|\\bin\\kwiver.exe” dump_klv video_file.mpeg**

and then at the end of the file, or probably in a file that gets included, we have

    .. |version| replace:: 1.1.0 

This way when we update the version we only have one place to change it.  Or possibly in the future we can have CMake write this file with the correct version number in it.

doc/manuals/installation.rst
```diff
@@ -0,0 +1,93 @@
+.. _installation:
+
+.. |telesculptor_install_1| image:: /images/telesculptor_install_1.png
+   :width: 49 %
```
mleotta, line 4: We don't need to force these to be half page width and make them fit two across.  That was just done to make them fit better on the page in Word.  Here they can be full size images (or maybe 80% width if full size looks too large) and they could be one on each line.  Fitting two on a line doesn't seem to work reliably in RST.

doc/manuals/telesculptor.rst
```diff
@@ -1,12 +1,58 @@
+.. _telesculptor:
```
mleotta, line 1: This file is a special case since it is currently what is built by CMake into a static webpage that is installed along side TeleSculptor.  That build currently uses `docutils` but not full blown sphinx processing so I don't think it will handle this new syntax correctly.  This file also has some content overlap with the rest of the manual.  I suggest we take this file out of the TOC and revert these changes for the initial pass.  This way the file will continue to work with the `docutils` build.  Once we have this initial documentation on readthedocs I would like to do the following:

1. Update the TeleSculptor build to use sphinx and build/install this new manual instead.
2.  Refactor the content from this file to better fit into the new manual.

These would both be after this current branch lands

mleotta: Note: you may want to cherry-pick the changes you made to this file onto a new branch.  Eventually we will need them, we just aren't ready for them yet.

doc/manuals/installation.rst
jacobderosa: Inline images with captions aren't permitted in RST. I was trying to find a workaround but none seem to exist.

mleotta: They don't need to be inline, you can do it this way.  My point was that they don't need to be 49% wide.  We could bump that up to may be 80% so each is on a new line.  The 49% sometimes puts image side-by-side but sometimes doesn't depending on the browser.  They don't need to be side-by-side in the HTML documentation so we don't need to force it.

doc/manuals/version.rst
```diff
@@ -0,0 +1,5 @@
+.. |version| unicode:: 1.1.0
+   :trim:
+
+.. |space| unicode:: 0xA0
```
mleotta, line 4: Why is it necessary to trim whitespace and then add it back in when needed?

If you don't use `:trim:` and write `v|version|` do you get `v1.1.0`, `v 1.1.0`, or does it just fail to replace without whitespace before the `|`?

jacobderosa: When inserting replacement text or unicode, you must put spaces on either side. v|version| results in "v|version|" because it's not recognized as an insertion. As a result, any inserted text has spaces on either side, which isn't always what we want, so I trimmed the whitespace and added it back when necessary.

mleotta: Got it.  Thanks

PR #471, opened 2021-05-07
only zoom to landmarks if landmarks have a valid size

### LINKED ISSUE #452: "Zoom to Landmarks" broken with no landmarks
The `Zoom to Landmarks` feature is currently broken when there are no landmarks.  The camera view is destroyed and there is no way to recover by zooming to all data.  If there are no landmarks, the zoom to landmarks feature should not do anything or possibly should be greyed out.

mleotta:
fixes #452

### DIFF
--- a/doc/release-notes/release.txt
+++ b/doc/release-notes/release.txt
@@ -1,0 +1,12 @@
+TeleSculptor v1.1.3 Release Notes
+=================================
+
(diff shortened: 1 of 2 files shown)


borovik135:
tested in my Ubuntu 20.04 and it seems to work OK now

mleotta:
I need to add release notes for this bug fix.

PR #482, opened 2021-05-18
Add vertex cells.

danlipsa:


danlipsa:
Solves
https://github.com/Kitware/TeleSculptor/projects/1#card-60652456

mleotta:
@borovik135 can you test this by loading a PLY that is just points with no faces?

borovik135:
I tested on KHQ point cloud and it works as intended: I was able to load the point cloud, mark/modify a few GCPs with their 2D registration points and successfully calibrate a camera; the point cloud got re-loaded on project re-load for the next session of TS.

PR #489, opened 2021-05-26
Replaced all headers with pointer to license file

quantumech3:
Replaced headers with a link to the license file for Telesculpter. Also removed the stray replace-copyright.sh script from maptk folder.


quantumech3:
Apologies I would like to test the build before this gets pushed. I forgot to do that.

mleotta (approved):

PR #491, opened 2021-06-01
Direct color RGBA and color map scalar.

danlipsa:
Also add vtp as possible input format.

### DIFF
--- a/gui/ColorizeSurfaceOptions.cxx
+++ b/gui/ColorizeSurfaceOptions.cxx
@@ -307,6 +307,7 @@ void ColorizeSurfaceOptions::colorize()
       {
         d->UI.comboBoxColorDisplay->setEnabled(true);
         emit colorModeChanged(d->UI.buttonGroup->checkedButton()->text());
+        this->InsideColorize = false;
         return;
       }
       QEventLoop loop;
--- a/gui/ColorizeSurfaceOptions.cxx
+++ b/gui/ColorizeSurfaceOptions.cxx
@@ -46,6 +46,13 @@ public:
   QString frameFile;
 
   kv::frame_id_t currentFrame = -1;
+
+  double occlusionThreshold = 1;
+  bool removeOccluded = true;
+  bool removeMasked = true;
+  bool insideColorize = false;
+  const int INVALID_FRAME = -2;
+  kwiver::vital::frame_id_t lastColorizedFrame = INVALID_FRAME;
 };
 
 QTE_IMPLEMENT_D_FUNC(ColorizeSurfaceOptions)
@@ -102,11 +109,6 @@ ColorizeSurfaceOptions::ColorizeSurfaceOptions(
   d->frameFile = QString();
 
   d->UI.comboBoxColorDisplay->setDuplicatesEnabled(false);
-  this->OcclusionThreshold = 1;
-  this->RemoveOccluded = true;
-  this->RemoveMasked = true;
-  this->InsideColorize = false;
-  this->LastColorizedFrame = INVALID_FRAME;
 }
 
 //-----------------------------------------------------------------------------
@@ -115,6 +117,34 @@ ColorizeSurfaceOptions::~ColorizeSurfaceOptions()
   QTE_D();
 
   d->uiState.save();
+}
+
+//-----------------------------------------------------------------------------
+void ColorizeSurfaceOptions::setOcclusionThreshold(double occlusionThreshold)
+{
+  QTE_D();
+  d->occlusionThreshold = occlusionThreshold;
+}
+
+//-----------------------------------------------------------------------------
+void ColorizeSurfaceOptions::setRemoveOccluded(double removeOccluded)
+{
+  QTE_D();
+  d->removeOccluded = removeOccluded;
+}
+
+//-----------------------------------------------------------------------------
+void ColorizeSurfaceOptions::setRemoveMasked(double removeMasked)
+{
+  QTE_D();
+  d->removeMasked = removeMasked;
+}
+
+//-----------------------------------------------------------------------------
+double ColorizeSurfaceOptions::getOcclusionThreshold()
+{
+  QTE_D();
+  return d->occlusionThreshold;
 }
 
 //-----------------------------------------------------------------------------
@@ -274,8 +304,6 @@ void ColorizeSurfaceOptions::changeColorDisplay()
   else
   {
     mapper->SetColorModeToDirectScalars();
-    mapper->CreateDefaultLookupTable();
-    mapper->UseLookupTableScalarRangeOff();
   }
 
   mapper->Update();
@@ -286,7 +314,8 @@ void ColorizeSurfaceOptions::changeColorDisplay()
 //-----------------------------------------------------------------------------
 void ColorizeSurfaceOptions::forceColorize()
 {
-  this->LastColorizedFrame = INVALID_FRAME;
+  QTE_D();
+  d->lastColorizedFrame = d->INVALID_FRAME;
   colorize();
 }
 
@@ -294,22 +323,29 @@ void ColorizeSurfaceOptions::forceColorize()
 void ColorizeSurfaceOptions::colorize()
 {
   QTE_D();
-  if (! this->InsideColorize)
+  std::cout << "InsideColorize: " << d->insideColorize << std::endl;
+  if (! d->insideColorize)
   {
-    this->InsideColorize = true;
+    d->insideColorize = true;
     auto colorizedFrame =
       (d->UI.radioButtonCurrentFrame->isChecked() ? d->currentFrame : -1);
-    while (this->LastColorizedFrame != colorizedFrame)
+    while (d->lastColorizedFrame != colorizedFrame)
     {
-      this->LastColorizedFrame = colorizedFrame;
+      std::cout << "begin while" << std::endl
+                << "LastColorizedFrame: " << d->lastColorizedFrame
+                << std::endl
+                << "colorizedFrame: " << colorizedFrame << std::endl;
+      d->lastColorizedFrame = colorizedFrame;
 
       if (!d->cameras || d->cameras->size() == 0)
       {
+        std::cout << "no cameras" << "\n\n";
         d->UI.comboBoxColorDisplay->setEnabled(true);
         emit colorModeChanged(d->UI.buttonGroup->checkedButton()->text());
-        this->InsideColorize = false;
+        d->insideColorize = false;
         return;
       }
+      std::cout << "setup coloration" << std::endl;
       QEventLoop loop;
       vtkPolyData* volume = vtkPolyData::SafeDownCast(d->volumeActor->GetMapper()
                                                       ->GetInput());
@@ -321,17 +357,17 @@ void ColorizeSurfaceOptions::colorize()
       coloration->set_input(volume);
       coloration->set_output(volume);
       coloration->set_frame_sampling(d->UI.spinBoxFrameSampling->value());
-      coloration->set_occlusion_threshold(this->OcclusionThreshold);
-      coloration->set_remove_occluded(this->RemoveOccluded);
-      coloration->set_remove_masked(this->RemoveMasked);
-      coloration->set_frame(this->LastColorizedFrame);
+      coloration->set_occlusion_threshold(d->occlusionThreshold);
+      coloration->set_remove_occluded(d->removeOccluded);
+      coloration->set_remove_masked(d->removeMasked);
+      coloration->set_frame(d->lastColorizedFrame);
       coloration->set_all_frames(false);
       connect(coloration, &MeshColoration::resultReady,
               this, &ColorizeSurfaceOptions::meshColorationHandleResult);
       connect( coloration, &MeshColoration::resultReady, &loop, &QEventLoop::quit );
       connect(coloration, &MeshColoration::finished,
               coloration, &MeshColoration::deleteLater);
-      if (this->LastColorizedFrame < 0)
+      if (d->lastColorizedFrame < 0)
       {
         // only enable the progress bar if coloring with multiple frames
         connect(coloration, &MeshColoration::progressChanged,
@@ -340,8 +376,12 @@ void ColorizeSurfaceOptions::colorize()
       coloration->start();
       loop.exec();
       colorizedFrame = (d->UI.radioButtonCurrentFrame->isChecked()) ? d->currentFrame : -1;
+      std::cout << "end while" << std::endl
+                << "LastColorizedFrame: " << d->lastColorizedFrame
+                << std::endl
+                << "colorizedFrame: " << colorizedFrame << std::endl;
     }
-    this->InsideColorize = false;
+    d->insideColorize = false;
   }
 }
 
--- a/gui/ColorizeSurfaceOptions.cxx
+++ b/gui/ColorizeSurfaceOptions.cxx
@@ -323,7 +323,6 @@ void ColorizeSurfaceOptions::forceColorize()
 void ColorizeSurfaceOptions::colorize()
 {
   QTE_D();
-  std::cout << "InsideColorize: " << d->insideColorize << std::endl;
   if (! d->insideColorize)
   {
     d->insideColorize = true;
@@ -331,21 +330,15 @@ void ColorizeSurfaceOptions::colorize()
       (d->UI.radioButtonCurrentFrame->isChecked() ? d->currentFrame : -1);
     while (d->lastColorizedFrame != colorizedFrame)
     {
-      std::cout << "begin while" << std::endl
-                << "LastColorizedFrame: " << d->lastColorizedFrame
-                << std::endl
-                << "colorizedFrame: " << colorizedFrame << std::endl;
       d->lastColorizedFrame = colorizedFrame;
 
       if (!d->cameras || d->cameras->size() == 0)
       {
-        std::cout << "no cameras" << "\n\n";
         d->UI.comboBoxColorDisplay->setEnabled(true);
         emit colorModeChanged(d->UI.buttonGroup->checkedButton()->text());
         d->insideColorize = false;
         return;
       }
-      std::cout << "setup coloration" << std::endl;
       QEventLoop loop;
       vtkPolyData* volume = vtkPolyData::SafeDownCast(d->volumeActor->GetMapper()
                                                       ->GetInput());
@@ -376,10 +369,6 @@ void ColorizeSurfaceOptions::colorize()
       coloration->start();
       loop.exec();
       colorizedFrame = (d->UI.radioButtonCurrentFrame->isChecked()) ? d->currentFrame : -1;
-      std::cout << "end while" << std::endl
-                << "LastColorizedFrame: " << d->lastColorizedFrame
-                << std::endl
-                << "colorizedFrame: " << colorizedFrame << std::endl;
     }
     d->insideColorize = false;
   }
--- a/gui/VolumeOptions.cxx
+++ b/gui/VolumeOptions.cxx
@@ -24,6 +24,8 @@
 
 #include <string>
 
+#include <vital/logger/logger.h>
+
 namespace // anonymous
 {
 
@@ -67,7 +69,11 @@ vtkSmartPointer<vtkColorTransferFunction> createBluesColorPalette(
 class VolumeOptionsPrivate
 {
 public:
-  VolumeOptionsPrivate():colorizeSurfaceOptions(nullptr),volumeActor(nullptr)
+  VolumeOptionsPrivate()
+    :  colorizeSurfaceOptions(nullptr),
+       volumeActor(nullptr),
+       originalColorArray(nullptr),
+       logger(kwiver::vital::get_logger("telesculptor.volumeoptions"))
   {}
 
   void setPopup(QToolButton* button, QWidget* popup);
@@ -78,7 +84,7 @@ public:
 
   vtkActor* volumeActor;
   vtkDataArray* originalColorArray;
-
+  kwiver::vital::logger_handle_t logger;
 };
 
 QTE_IMPLEMENT_D_FUNC(VolumeOptions)
@@ -111,7 +117,6 @@ VolumeOptions::VolumeOptions(const QString &settingsGroup, QWidget* parent,
   d->uiState.restore();
 
   d->colorizeSurfaceOptions = new ColorizeSurfaceOptions(settingsGroup, this);
-  d->originalColorArray = nullptr;
   d->setPopup(d->UI.toolButtonColorizeSurfaceMenu, d->colorizeSurfaceOptions);
 
   // Connect signals/slots
@@ -259,10 +264,19 @@ void VolumeOptions::setCurrentFrame(kwiver::vital::frame_id_t frame)
 }
 
 //-----------------------------------------------------------------------------
-void VolumeOptions::setSurfaceColor(int index)
+void VolumeOptions::setColorizeSurface(int index)
 {
   QTE_D();
 
+  if (index < 0 || index > ORIGINAL_COLOR)
+  {
+    LOG_ERROR(d->logger, "Invalid index: " << index);
+  }
+  if (index == ORIGINAL_COLOR &&
+      d->UI.comboBoxColorizeSurface->count() <= 2)
+  {
+    d->UI.comboBoxColorizeSurface->insertItem(index, "Original color");
+  }
   d->UI.comboBoxColorizeSurface->setCurrentIndex(index);
 }
 
--- a/gui/VolumeOptions.cxx
+++ b/gui/VolumeOptions.cxx
@@ -16,12 +16,52 @@
 #include <QWidgetAction>
 
 #include <vtkActor.h>
+#include <vtkColorTransferFunction.h>
 #include <vtkPolyDataMapper.h>
 #include <vtkDataSet.h>
 #include <vtkPointData.h>
 #include <vtkPolyData.h>
 
 #include <string>
+
+namespace // anonymous
+{
+
+//-----------------------------------------------------------------------------
+vtkSmartPointer<vtkColorTransferFunction> createBluesColorPalette(
+  double* range)
+{
+  auto lut = vtkSmartPointer<vtkColorTransferFunction>::New();
+  std::array<std::array<double, 3>, 17> blues = {{
+      {0.031373, 0.188235, 0.419608},
+      {0.031373, 0.253195, 0.516063},
+      {0.031757, 0.318139, 0.612149},
+      {0.080969, 0.38113, 0.661361},
+      {0.130427, 0.444152, 0.710327},
+      {0.195386, 0.509112, 0.743791},
+      {0.260715, 0.573841, 0.777209},
+      {0.341423, 0.628958, 0.808704},
+      {0.422745, 0.684075, 0.839892},
+      {0.523137, 0.739193, 0.861546},
+      {0.622684, 0.793464, 0.883429},
+      {0.701423, 0.826928, 0.910988},
+      {0.778685, 0.8603, 0.937993},
+      {0.825928, 0.891795, 0.953741},
+      {0.87328, 0.923291, 0.969489},
+      {0.922491, 0.954787, 0.985236},
+      {0.968627, 0.984314, 1}}};
+  for (size_t i = 0; i < blues.size(); ++i)
+  {
+    lut->AddRGBPoint(
+      range[0] + (range[1] - range[0]) * i / (blues.size() - 1),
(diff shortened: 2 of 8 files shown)


danlipsa:
@mleotta @borovik135 @mwoehlke-kitware Please review.

borovik135:
I have tested this PR with a mesh that contains an alpha channel, which loads and displays fine: all colors visible as expected.

gui/WorldView.cxx
```diff
@@ -85,10 +87,47 @@
...
 
+#include <array>
+
 using namespace LandmarkArrays;
 
 QTE_IMPLEMENT_D_FUNC(WorldView)
 
+
+namespace {
+vtkSmartPointer<vtkColorTransferFunction> setupBluesColorPalette(double* range)
```
mwoehlke-kitware, line 98:
```suggestion
namespace // anonymous
{

//-----------------------------------------------------------------------------
vtkSmartPointer<vtkColorTransferFunction> setupBluesColorPalette(double* range)
```

gui/WorldView.cxx
```diff
@@ -85,10 +87,47 @@
...
+  {
+    lut->AddRGBPoint(
+      range[0] + (range[1] - range[0]) * i / (blues.size() - 1),
+      blues[i][0], blues[i][1], blues[i][2]);
+  }
+  // blues from ParaView: ColorMap.json
+  lut->SetColorSpaceToLab();
+  return lut;
+}
+}
```
mwoehlke-kitware, line 129:
```suggestion

} // namespace <anonymous>
```

gui/WorldView.cxx
```diff
@@ -85,10 +87,47 @@
...
 using namespace LandmarkArrays;
 
 QTE_IMPLEMENT_D_FUNC(WorldView)
 
+
+namespace {
+vtkSmartPointer<vtkColorTransferFunction> setupBluesColorPalette(double* range)
+{
+  auto lut = vtkSmartPointer<vtkColorTransferFunction>::New();
+  std::array<std::array<double, 3>, 17> blues = {{
```
mwoehlke-kitware, line 101: `std::array` almost seems like overkill here, especially as it forces declaring the number of colors up front.

BTW... this is to colorize points with only one channel?

gui/WorldView.cxx
```diff
@@ -85,10 +87,47 @@
...
+      {0.422745, 0.684075, 0.839892},
+      {0.523137, 0.739193, 0.861546},
+      {0.622684, 0.793464, 0.883429},
+      {0.701423, 0.826928, 0.910988},
+      {0.778685, 0.8603, 0.937993},
+      {0.825928, 0.891795, 0.953741},
+      {0.87328, 0.923291, 0.969489},
+      {0.922491, 0.954787, 0.985236},
+      {0.968627, 0.984314, 1}}};
+  for (size_t i = 0; i < blues.size(); ++i)
```
mwoehlke-kitware, line 119:
```suggestion
  for (auto const i : qtIndexRange(blues.size()))
```

(Note: `qtIndexRange` and `kwiver::vital::range::iota` are the same thing, but TS currently has several uses of the former and none of the latter, and `qtIndexRange.h` is already included.)

gui/WorldView.cxx
```diff
@@ -979,27 +1018,40 @@ void WorldView::computeContour(double threshold)
...
-  vtkNew<vtkPLYReader> reader;
-  reader->SetFileName(qPrintable(path));
-  reader->Update();
-  QGuiApplication::restoreOverrideCursor();
-
-  auto mesh = vtkPolyData::SafeDownCast(reader->GetOutput());
-  if (mesh->GetPointData()->GetNumberOfArrays() <= 0)
+  QFileInfo info(path);
+  vtkSmartPointer<vtkPolyData> mesh;
+  QString ext = info.suffix().toLower();
```
mwoehlke-kitware, line 1032:
```suggestion
  vtkSmartPointer<vtkPolyData> mesh;
  auto const& ext = fileInfo.suffix().toLower();
```

gui/WorldView.cxx
```diff
@@ -979,27 +1018,40 @@ void WorldView::computeContour(double threshold)
...
   QGuiApplication::setOverrideCursor(Qt::WaitCursor);
-
-  // Create the vtk pipeline and read the mesh
-  vtkNew<vtkPLYReader> reader;
-  reader->SetFileName(qPrintable(path));
-  reader->Update();
-  QGuiApplication::restoreOverrideCursor();
-
-  auto mesh = vtkPolyData::SafeDownCast(reader->GetOutput());
-  if (mesh->GetPointData()->GetNumberOfArrays() <= 0)
```
mwoehlke-kitware, line 997: Maybe we shouldn't lose this check?

gui/WorldView.cxx
```diff
@@ -1035,13 +1087,40 @@ void WorldView::setMesh(vtkSmartPointer<vtkPolyData> mesh)
...
   {
+    rgb = true;
+    arrayName = rgbName;
+  }
+  if (mesh->GetPointData()->HasArray(rgbaName))
+  {
+    rgba = true;
+    arrayName = rgbaName;
+  }
+  if (rgb || rgba)
```
mwoehlke-kitware, line 1104:
```suggestion
  if (arrayName)
```
...and lose `rgb`, `rgba`.

gui/WorldView.cxx
```diff
@@ -1035,13 +1087,40 @@ void WorldView::setMesh(vtkSmartPointer<vtkPolyData> mesh)
 
   // Add this actor to the renderer
   d->renderer->AddActor(d->volumeActor);
-
-  if (mesh->GetPointData()->HasArray("RGB"))
+  bool rgb = false, rgba = false;
+  const char* arrayName = nullptr;
+  const char* rgbName = "RGB";
+  const char* rgbaName = "RGBA";
```
mwoehlke-kitware, line 1093:
```suggestion
  char const* arrayName = nullptr;
  constexpr auto rgbName = "RGB";
  constexpr auto rgbaName = "RGBA";
```

Please ***always*** write `T const`, never `const T`. I believe most Vision projects are following this these days, but definitely KWIVER, Krest and TeleSculptor are.

gui/WorldView.cxx
```diff
@@ -1035,13 +1087,40 @@ void WorldView::setMesh(vtkSmartPointer<vtkPolyData> mesh)
...
+  if (mesh->GetPointData()->HasArray(rgbName))
   {
+    rgb = true;
+    arrayName = rgbName;
+  }
+  if (mesh->GetPointData()->HasArray(rgbaName))
```
mwoehlke-kitware, line 1099:
```suggestion
  else if (mesh->GetPointData()->HasArray(rgbaName))
```

Is it possible for a mesh to have both? I would think not, but you may or may not also want to check for RGBA first.

gui/WorldView.cxx
mwoehlke-kitware, line 98:
```suggestion
vtkSmartPointer<vtkColorTransferFunction> createBluesColorPalette(
  double* range)
```

gui/WorldView.cxx
```diff
@@ -1035,13 +1087,40 @@ void WorldView::setMesh(vtkSmartPointer<vtkPolyData> mesh)
...
     meshMapper->ScalarVisibilityOn();
-    meshMapper->SelectColorArray("RGB");
     d->volumeOptions->setSurfaceColored(true);
   }
+  else if (mesh->GetPointData()->GetScalars())
+  {
+    // map active scalar
+    vtkDataArray* scalars = mesh->GetPointData()->GetScalars();
+    double range[2];
+    scalars->GetRange(range);
```
mwoehlke-kitware, line 1116:
```suggestion
    double range[2];
    mesh->GetPointData()->GetScalars()->GetRange(range);
```

gui/WorldView.cxx
```diff
@@ -1035,13 +1087,40 @@ void WorldView::setMesh(vtkSmartPointer<vtkPolyData> mesh)
...
+    auto lut = ::setupBluesColorPalette(range);
+    meshMapper->SetLookupTable(lut);
```
mwoehlke-kitware, line 1118:
```suggestion
    meshMapper->SetLookupTable(createBluesColorPalette(range));
```

mwoehlke-kitware (changes requested):
Looks functionally okay (mostly¹), but please fix at least the more severe style regressions.

(¹ See note on losing the check that the imported mesh is non-empty.)

gui/WorldView.cxx
danlipsa: Yes.

gui/WorldView.cxx
danlipsa: This does not bring in much: it's longer and uses qtIndexRange which I never seen so its not necessarily clearer. So I would not do this.

gui/WorldView.cxx
danlipsa: I am not a big fan of auto in these simpler cases where it does not even save space but make the type less clear. You have to lookup the return of toLower to see that the type is. But I will use the project const placement.

gui/WorldView.cxx
danlipsa: Sounds good.

gui/WorldView.cxx
danlipsa: The way its implemented, it checks for RGBA last so it uses that for coloring if a dataset has both RGB and RGBA. I think that is reasonable. We can refine this in a future PR and maybe give the user a choice.

gui/WorldView.cxx
danlipsa: Sounds good.

gui/WorldView.cxx
danlipsa: Why is this better?

danlipsa: Oh, scalars is not used anywhere else. Sounds good.

borovik135:
checked the updated functionality with the RGBA mesh supplied by DL, and it appears to work as expected with visible semi-transparency, tnx

gui/WorldView.cxx
mwoehlke-kitware: It makes the value immutable. It makes it so you don't need to know the correct type (which, yes, *is* an issue because there are classes floating around that use `int` for indexing.)

Would you, in Python, write a `for` loop like this? Or would you write `for i in range(blues.size())`? Would you decline to use that because "I've never seen [range]"?

gui/WorldView.cxx
mwoehlke-kitware: The way it's implemented is premature pessimization. (Note, I suggested also switching the checks for RGB and RGBA.) Always checking for RGB and then potentially discarding that work is silly.

gui/WorldView.cxx
mwoehlke-kitware: Using `const` would be good. The "biggest" change here though is not gratuitously recreating the `QFileInfo`.

gui/WorldView.cxx
danlipsa: Oh, I see. It makes it Python like. Didn't realize this initially. Still don't see why would you do this when there is a standard way to do this in C/C++.

gui/WorldView.cxx
danlipsa: Sounds good.

gui/WorldView.cxx
danlipsa: Sounds good.

gui/WorldView.cxx
mwoehlke-kitware: The "standard way to do this in C/C++" (well, in C++ anyway) is [`std::views::iota`](https://en.cppreference.com/w/cpp/ranges/iota_view), which a) unfortunately isn't available yet¹ and b) unfortunately requires specifying the starting index. And possibly also requires specifying the type explicitly. (I'm not entirely clear why a more direct equivalent hasn't been standardized, given how frequently it's needed.)

(¹ Requires C++20, which we are some time from being able to use.)

gui/WorldView.cxx
danlipsa: While more efficient, it is slightly less clear so I am on the fence here.

borovik135: if we ultimately decide for arrayName, then we probably do not need bool rgb, rgba; unless they get some usage later on; just my 2c

mwoehlke-kitware: Hmm, I might have argued the opposite: "if I found an array to use for coloration" vs. "if I found either RGB or RGBA". Sort of like `if (haveSufficientFunds)` vs `if (bankBalance > costOfItem)`. The latter is "clearer" in the sense of explicitly restating the actual conditions, while the former places more emphasis on the *intent*. I suppose I can see the argument either way (but one is less code 😉).

gui/WorldView.cxx
mwoehlke-kitware: (Note: as discussed elsewhere, this wasn't doing exactly what I thought it was doing. We might want to revisit checking that we actually loaded data, though. Not necessarily in this PR.)

gui/WorldView.cxx
danlipsa: Sounds good.

danlipsa:
One thing I forgot to mention is that with this PR I added VTP in the list of supported formats for meshes besides PLY. This is because the PLY writer in VTK is not very performant. It did not allow me to save a mesh with one scalar or an RGBA attribute correctly.

gui/WorldView.cxx
danlipsa: The humble for loop to the rescue 😄

borovik135:
I played with different mesh formats in TS. This PR indeed imports/exports the VTP format fine. However, RGBA mesh imported from VTP will not get exported to PLY or LAS producing a core dump. RGBA export to OBJ results in a non-colored mesh. I suggest taking a closer look at TS mesh export function, as I suspect it may miss a check if the mesh may be RGBA.

danlipsa:
The reason I added VTP is that the PLY writer in VTK didn't work correctly - I could not save a mesh with one scalar. I don't know about the OBJ. LAS export does not use VTK so I don't know how that is handled. Is that suppose to only save the points? I could take a look at this in this PR or maybe a different one? @mleotta ?

borovik135:
In any case, the program should not crash producing a core dump.

danlipsa:
I agree. Its strange that it crashes.

danlipsa:
@borovik135 @mleotta @mwoehlke-kitware It should not crash when saving a transparent mesh and it should color an imported mesh.

borovik135:
Export fused mesh does not crash on RGB or RGBA colored examples I have tried.

borovik135:
Camera based mesh coloring appears to function for computed cameras and imported meshes, but in a strange way: a computed camera image projection on an already colored mesh is not apparent unless one clears 'Colorize surface' getting a blank mesh and then sets 'Colorize surface' back on, our computed camera image projected onto the mesh, but with artifacts that are not supposed to be there, ignoring mesh self-occlusions. Getting back to the imported fully colored mesh does not appear to be possible without reloading the project.
Screenshot from 2021-06-09 11-19-27

borovik135:
Mesh coloring for a couple of computed/calibrated cameras looks even more confusing. Am I doing something wrong? Is there an option to account for the mesh occlusions?
Screenshot from 2021-06-09 11-21-38

borovik135:
I think Computed Camera image back-projection on the mesh should emulate an image projector on a colored surface accounting for occlusions and mixing its image colors with the mesh colors, preferably allowing switching/blending back and forth between the image and mesh colors. Is that possible/feasible given the project constraints?

mleotta:
> Mesh coloring for a couple of computed/calibrated cameras looks even more confusing. Am I doing something wrong? Is there an option to account for the mesh occlusions?
Yes, there should be an option in the mesh color
image

borovik135:
changing occlusion threshold values does not seem to have any effect on coloring

mleotta:
Recovering the original color is a new feature request outside of this issue, but occlusion not working is a regression.  @danlipsa is the occlusion still working for you?  Maybe there is some connection between this occlusion regression and the speed regression?

borovik135:
On my system image based mesh coloring happens very quickly, yet I am using a fairly coarse KHQ mesh.

gui/ColorizeSurfaceOptions.cxx
```diff
@@ -307,6 +307,7 @@ void ColorizeSurfaceOptions::colorize()
       {
         d->UI.comboBoxColorDisplay->setEnabled(true);
         emit colorModeChanged(d->UI.buttonGroup->checkedButton()->text());
+        this->InsideColorize = false;
```
mwoehlke-kitware, line 310: Somewhat OT as this existed previously, but why are there a bunch of protected members that don't follow the naming convention when this class has a PIMPL?

gui/WorldView.cxx
```diff
@@ -85,10 +87,107 @@
...
+  }
+  // blues from ParaView: ColorMap.json
+  lut->SetColorSpaceToLab();
+  return lut;
+}
+
+//-----------------------------------------------------------------------------
+bool validForColoring(vtkDataArray* a, bool& mapScalars)
+{
+  if (! a)
```
mwoehlke-kitware, line 137:
```suggestion
  if (!a)
```

gui/WorldView.cxx
```diff
@@ -1035,15 +1146,26 @@ void WorldView::setMesh(vtkSmartPointer<vtkPolyData> mesh)
...
+      meshMapper->SetColorModeToMapScalars();
+    }
+    else
+    {
+      // for char, short, int, long arrays 0-255 maps to color component
+      // for float arrays 0-1 maps to color component
+      meshMapper->SetColorModeToDirectScalars();      
+    }
     d->volumeOptions->setSurfaceColored(true);
+    emit contourChanged();
```
mwoehlke-kitware, line 1167: Why is `countourChanged` only emitted if there are scalars?

danlipsa: Good catch. It needs to be emitted always.

danlipsa:
> Recovering the original color is a new feature request outside of this issue, 

Looking at this it would have been as difficult and a lot more confusing to make the coloring work for additional arrays the first time and then not work any more. So the previous checkbox: Colorize Surface (check, uncheck) became a combobox:
(No color, Original color, Image color). So Colorize surface uncheck corresponds to No color.
Colorize surface check corresponds to Image color. Original color is a new possibility.


> but occlusion not working is a regression. @danlipsa is the occlusion still working for you?

Yes, occlusion works fine on the 09172008flight1tape3_2 example.

 > Maybe there is some connection between this occlusion regression and the speed regression?

Still haven't look at the speed issue. More details: It takes about 50 sec for a mesh with 4.7 millions cells. I down-sampled the mesh and worked with one with 0.47 million cells and that takes about 7 sec. If you don't see any difference on the examples you  have used in the past maybe I just generated a bigger mesh.

danlipsa:
@borovik135 @mwoehlke-kitware @mleotta Please review.

mleotta:
The concept of "Original Color" does not make sense for the "normal" use case of TeleSculptor, which is to produce a mesh from the images.  Is there any easy way to make the GUI only show that "Original Color" option if the mesh was loaded from a file that had color data in it?

danlipsa:
> The concept of "Original Color" does not make sense for the "normal" use case of TeleSculptor, which is to produce a mesh from the images. Is there any easy way to make the GUI only show that "Original Color" option if the mesh was loaded from a file that had color data in it?

Good idea. Yes, it should not be too difficult. I'll take a look.

danlipsa:
One problem (which existed before) with this approach which was fixed by the "Original color" is that a mesh in TeleSculptor is loaded from a PLY which you open a project. This means that only color is loaded, so no mean, median or count are available only RGB. I can save as VTP instead of PLY which I think will help. @mleotta What do you think?

mleotta:
If/when we **automatically** save the mesh **into the project directory** so that it is reloaded when the project is opened, we should probably use VTP and store the multiple color channels.  This way we can recover the state as it originally was.  However, a user should still be able to File->export->mesh and save the mesh as a PLY with only the active color stored as RGB.  They could also choose to export the mesh as VTP.

I don't recall if we ever implemented automatic saving of the mesh to the project directory, but we should, and that should use VTP.  However, I don't want this branch to keep ballooning to add more features.  It would be best to get this landed with only minimal changes and then make a new PR to address further improvements.

danlipsa:
OK, Sounds good. For now I fixed this by only setting IMAGE_COLOR when opening a project if the data created has mean, median and count. Otherwise I set NO_COLOR - this always happens now as the data is read from a PLY.

mleotta:
@danlipsa this recent change isn't quite right.  The behavior shouldn't change depending on whether the mesh was manually imported or loaded from a project file.  In either case, if the mesh has required color data (mean, median, etc.) we should restore that data so it appears as it does when computed.  Currently this would be the case if you load a VTP saved by TeleSculptor.  If the mesh does not have all the original color data, but does have some color or grayscale data (e.g. RGB) then we should put this in ORIGINAL_DATA.  This would be the case for PLY files from TeleSculptor or from some other source.

The point I was making previously relates to how **exporting** should work, not importing.  If we automatically save the mesh data to the project when it is changed we should do this as VTP, but when the user manually exports a mesh they get to pick the format.  It's possible a user could manually edit the project file to put a PLY in the project, so we shouldn't assume the project will only contain VTP files.

danlipsa:
Sounds good.
What you are describing sounds like we don't use the last commit.
Instead, we only add ORIGINAL_COLOR if the mesh does not have the required arrays (mean, median and count).

mleotta:
yes, that sounds right.

borovik135:
KHQ frames and model to test the occlusion problem are located at smb://videonas/other/data_working/KHQ_Construction/2017-07-14/GOPR0369

danlipsa:
Thanks Eugene. I'll try it out.

danlipsa:
@borovik135 How do you compute a mesh out of those images? I did New Project and then File / Inport / Imagery ...
but All supported files does not show me anything. Thanks.

borovik135:
The typical workflow for camera calibration is to load frames (e.g. frames.txt), import a pre-computed mesh (e.g. 3D/building.parking.coarse.mesh.ply) and possibly some pre-clicked GCP/CRPs (e.g. from 3D/20190714_frm_gcps.json), compute a camera and observe image projection on the mesh.

danlipsa:
When I run Batch Compute Depth Maps
I got
<pre>
DEBUG - reading image from file "/home/danlipsa/data/TeleSculptor/GOPR0369/frames/GOPR0369-06237.png"
TeleSculptor: /home/danlipsa/projects/TeleSculptor/build/external/fletch-build/install/include/vxl/core/vil/vil_crop.h:25: vil_image_view<T> vil_crop(const vil_image_view<T>&, unsigned int, unsigned int, unsigned int, unsigned int) [with T = double]: Assertion `j0<im.nj()' failed.
Aborted (core dumped)
[~/projects/TeleSculptor/build/external/telesculptor-build]$ 
</pre>
Have you seen this before?

danlipsa:
> compute a camera and observe image projection on the mesh.
Can you tell me how do I compute a camera after I load the GCPs? Thanks!

borovik135:
To compute a camera after importing GCPs, one needs to pick a frame that has corresponding camera registration points (CRPs), e.g. by expanding a tree branch  in the GCP list and double-clicking some frame number and then choosing compute camera drop-down (under the target-like button) in the camera view.

mleotta:
@danlipsa there should be no need to compute depth maps on the KHQ data from @borovik135.  TeleSculptor is not working on that data yet.  You should just be loading the project that @borovik135 provided including precomputed camera models, images, and 3D mesh.  @borovik135 can you make sure @danlipsa has some cameras for that data too?

borovik135:
posted some calibrated cameras to smb://videonas/other/data_working/KHQ_Construction/2017-07-14/GOPR0369/3D

danlipsa:
For me, the mesh is much bigger than the grid + the cameras. The first picture shows the grid and cameras. The second picture shows the grid, cameras and the mesh colored with the original color - not the grid completely covers the grid (the second picture is at the same scale as the first picture). The third picture shows the grid colored using the images for all cameras, at the same scale but rotated. What am I doing wrong?
grid-cameras
grid-cameras-originalcolor
grid-cameras-imagecolor-allframes

danlipsa:
@borovik135 @mleotta Any suggestions for my previous post? Thanks!

borovik135:
@danlipsa, I suggest skipping any feature or volume computation. Open TS, load frames (frames.txt), import a pre-computed mesh (3D/building.parking.coarse.mesh.ply) and some GCP/CRPs (3D/20190714_frm_gcps.json). Jump to a frame with CRPs (116, 320, 368) compute a camera [
image
]
Screenshot from 2021-06-17 13-55-46
 and observe image projection on the mesh.

borovik135:
@danlipsa, feel free to schedule a meeting so I can walk you through the suggested steps.

mleotta:
@borovik135 can you just save a project that has the cameras you computed, the mesh, and the images altogether so that Dan can just load it and doesn't need to learn how to do this or duplicate your work?  We are just trying to quickly get @danlipsa to be able to reproduce the issue with occlusion in mesh coloring.

danlipsa:
Seems that my version of telesculptor does not have the camera computation - there is no down arrow like in the image you show. Let me update and rebuild.

danlipsa:
I am all set. Eugene showed me what I was doing wrong.

mleotta (approved):
This looks good to me.  Any further improvements or fixes should happen on a new PR.  This PR already goes well beyond the feature it was supposed to be adding.

gui/VolumeOptions.cxx
```diff
@@ -216,44 +264,124 @@ void VolumeOptions::setCurrentFrame(kwiver::vital::frame_id_t frame)
 }
 
 //-----------------------------------------------------------------------------
-void VolumeOptions::setSurfaceColored(bool enabled)
+void VolumeOptions::setColorizeSurface(int index, bool blockSignals)
```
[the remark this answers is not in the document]
danlipsa: there is a showColorizeSurfaceMenu which gets called by setColorizeSurface. I rename the function to match. But I agree, the name is not very good. I think the name came from a checkbox called Colorize Surface.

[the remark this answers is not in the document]
danlipsa: Sounds good.

gui/VolumeOptions.cxx
```diff
@@ -216,44 +264,124 @@ void VolumeOptions::setCurrentFrame(kwiver::vital::frame_id_t frame)
...
 {
   QTE_D();
 
-  d->UI.checkBoxColorizeSurface->setChecked(enabled);
+  if (index < 0 || index > ORIGINAL_COLOR)
+  {
+    LOG_ERROR(d->logger, "Invalid index: " << index);
```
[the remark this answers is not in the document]
danlipsa: Sounds good.

gui/VolumeOptions.cxx
```diff
@@ -216,44 +264,124 @@ void VolumeOptions::setCurrentFrame(kwiver::vital::frame_id_t frame)
...
 
-  return d->UI.checkBoxColorizeSurface->isCheckable()
-         && d->UI.checkBoxColorizeSurface->isChecked();
+  return  d->UI.comboBoxColorizeSurface->currentIndex() == IMAGE_COLOR;
 }
 
+
 //-----------------------------------------------------------------------------
-void VolumeOptions::showColorizeSurfaceMenu(bool state)
+bool VolumeOptions::validForColoring(vtkDataArray* a, bool& mapScalars)
```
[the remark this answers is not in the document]
danlipsa: Sounds good.

danlipsa:
If this is ready, let's merge it.

danlipsa:
@borovik135 I notice that the image projection on the mesh is not properly aligned. Do you see the same thing?
telesculptor

borovik135:
My projections appear to be aligned fairly well. @danlipsa notice that the camera registration points (CRPs) appear to be significantly off in your camera view: they do not seem to correspond to the ground control points. It looks like CRPs were either modified or the. I have posted a project conf file to smb://videonas/other/data_working/KHQ_Construction/2017-07-14/GOPR0369; try using that; you may need to adjust some file paths for your local system. We can briefly meet to make sure the needed items are loaded and used properly.
Screenshot from 2021-06-23 13-16-01

mleotta:
Is this possibly loading the wrong images, or in the wrong order?  @danlipsa make sure you have same image list text file as @borovik135, modulo changes in file paths needed to find the images.  The CRPs are assigned to the image based on frame number (order in the file) not based on file name.

danlipsa:
@mleotta Indeed it was the frames image files that were wrong. @borovik135 helped me get the right ones and now it is aligned.

PR #500, opened 2021-06-26
Allow occlusion shift in both directions.

danlipsa:


danlipsa:
@mleotta @borovik135 Please review.

mleotta (commented):
I don't understand why we would need to use a negative threshold here.

borovik135:
negative values cause more black-out regions, positive values cause fewer black-out regions, but more of the ghost see-through kind of shadows, e.g. as with the 
Screenshot from 2021-06-28 13-16-35
AC units on the roof of the building

danlipsa:
@mleotta Should we merge?

mleotta:
I still don't understand why negative threshold values give you any different result than zero, but I suppose this change can't hurt.

danlipsa:
See https://github.com/Kitware/kwiver/blob/dee069057a088f3d537637968dd19943480a1828/arrows/vtk/mesh_coloration.cxx#L434
where we compare the depth from the depth buffer with the depth computed in software.
If a point is not occluded the depthbuffer depth should be equal with software depth. If a point is occluded, the software depth should be smaller. Plus and minus threshold adjust the two depths so that the equality happens on the right points.

PR #510, opened 2021-08-20
Remove maptk tools

chetnieter:
Removing deprecated maptk tools and references to them .


README.rst
```diff
@@ -324,92 +324,6 @@ in the process of being migrated to KWIVER and will leave this repository soon.
 Continued support is not guaranteed and behavior may diverge from documentation.
```
mleotta, line 324: We should remove this paragraph as well.  And the section heading above it.  and remove tools from the directory list on line 315.

doc/release-notes/master.txt
```diff
@@ -10,6 +10,9 @@ Updates since v1.1.0
 
 TeleSculptor Application
 
+ * Removed deprecated command line maptk tools. All the underlying kwiver
+   algoritms can be accessed via the 'kwiver' command line tool.
```
mleotta, line 14: I would put this comment in a new section outside of "TeleSculptor Application".  Probably under a "Tools" heading.

mleotta (changes requested):
There are other places in the README that refer to the tools or tools directory or tools cmake option.  Search for all instances of "tool".

PR #515, opened 2021-08-23
Update 3rd party library versions

mleotta:


mleotta:
I have confirmed that this builds successfully and runs as expected on Window.  @chetnieter , can you confirm for Linux and Mac?

chetnieter:
> I have confirmed that this builds successfully and runs as expected on Window. @chetnieter , can you confirm for Linux and Mac?

It builds for me on MacOS. I was having some issues with Qt building on Linux but I think I was missing some x11 development libraries. I try again this afternoon.

PR #523, opened 2021-09-06
Fixed a crash when fusing depth after a mesh has been loaded from disk

mleotta:
We need to switch back to "no color" and disable the "original color" option
when a fused volume mesh replaces a mesh that was loaded from a file.


mleotta:
@danlipsa This PR fixes the issue we talked about.  If a mesh is loaded and then we go back to a mesh extracted from a volume we need to switch back to "no color" mode and remove the "original color" option again.  Does this seem like a reasonable fix to you?

danlipsa:
Looks good. +1

PR #525, opened 2021-09-06
Rename compute "option" to "settings"

mleotta:
This works around a Qt bug on MacOS where menu items can not be named "options".

I've also updated the naming in the manual and provided an updated screenshot


chetnieter:
@mleotta Have you tested this on your mac? When I build and install it on my mac I still do not see the Settings sub-menu under compute.

mleotta:
hmmm.  I tested the name change on my Mac and found that most other string name showed up, including "Settings", but "Options" did not show up.   Apparently there are special names like "Options" and "Preferences" that Qt tries to move up the application name menu on MacOS.  It seems if that item is a submenu it just gets lost rather than moved.

I'm not sure why "Settings" worked for me but not for you.

chetnieter:
> hmmm. I tested the name change on my Mac and found that most other string name showed up, including "Settings", but "Options" did not show up. Apparently there are special names like "Options" and "Preferences" that Qt tries to move up the application name menu on MacOS. It seems if that item is a submenu it just gets lost rather than moved.
> 
> I'm not sure why "Settings" worked for me but not for you.

Could you put your installer on videonas and I will try it out on my machine.

mleotta:
You are right.  I built it again and it didn't work.  Apparently "Settings" is also a reserved name.  I can't find anyway to bypass this bug other than to make the name "*Settings" or some other string that doesn't exactly match settings, options, preferences, config, etc.

https://doc.qt.io/qt-5/qmenubar.html#qmenubar-as-a-global-menu-bar

mleotta:
I'm going to break out the doc updates that we need anyway into a new PR and revert the change from "Options" to "Settings", but we still need to fix to make "Options" work on MacOS.

@mwoehlke-kitware do you have any idea how to work around Qt relocating (or dropping) menu items named "Options" on MacOS?  The docs suggest setting `QAction::menuRole()` to override this behavior, but I don't know how to do that here because the item in question is a `QMenu`, not a `QAction`, so I can't set the role.  It seems like this might be a Qt bug that it is removing `QMenu` items matching "Options" and not placing them in the application menu.

mleotta:
I moved the doc updates without the renaming to #527.  The MacOS bug still needs to be addressed.  I'm going to close this PR, but we need a different fix

mleotta:
Opened #528 so we don't lose track of this issue.

PR #530, opened 2021-09-13
Added a hack to show the Options menu on MacOS

mleotta:
This is an ugly hack, but I'm not sure of a better way to make this work.
If on Apple then replace the menu title "Options" with "0ptions" (note the zero instead of O).
This seems to work and visually not that noticeable, but we should find a better fix at some point.

### DIFF
--- a/gui/MainWindow.cxx
+++ b/gui/MainWindow.cxx
@@ -1499,6 +1499,11 @@ MainWindow::MainWindow(QWidget* parent, Qt::WindowFlags flags)
     std::bind(&MainWindowPrivate::handleLogMessage, d, _1, _2, _3, _4);
   kv::kwiver_logger::set_global_callback(cb);
 
+  // Work around issue where "Options" menu will not display on MacOS
+#ifdef __APPLE__
+  d->UI.menuComputeOptions->setTitle("0ptions");
+#endif
+
   d->toolMenu = d->UI.menuCompute;
   d->toolSeparator =
     d->UI.menuCompute->insertSeparator(d->UI.actionCancelComputation);


gui/MainWindow.cxx
```diff
@@ -1499,6 +1499,11 @@ MainWindow::MainWindow(QWidget* parent, Qt::WindowFlags flags)
     std::bind(&MainWindowPrivate::handleLogMessage, d, _1, _2, _3, _4);
   kv::kwiver_logger::set_global_callback(cb);
 
+  // Work around issue where "Options" menu will not display on MacOS
+#ifdef __APPLE__
+  d->UI.menuComputeOptions->setTitle("0ptions");
```
mwoehlke-kitware, line 1504: This is likely to also break the ability to select the menu using the keyboard. If the issue is an exact string match being performed, I would strongly recommend at least *trying* inserting a non-printing character such as a [ZWSP](https://en.wikipedia.org/wiki/Zero-width_space):
```suggestion
  d->UI.menuComputeOptions->setTitle(QStringLiteral(u"Options\u200b"));
```

Other things to try if that doesn't work: stick the ZWSP in the middle rather than the end, or try [WJ](https://en.wikipedia.org/wiki/Word_joiner) (`\u2060`).

mleotta: I tried several such spaces in different positions.  I couldn't get it to work.  I'm really not that worried about the keyboard shortcut on this menu item being broken on Apple.  That's much better than the menu being missing entirely.

mwoehlke-kitware: Sure, but *not* having the visual "glitch" is surely better.
What about:
```suggestion
  d->UI.menuComputeOptions->setTitle(QStringLiteral(u"Opt\u2170ons"));
```

chetnieter: I will give it a try this afternoon.

chetnieter: This latest suggestion does not work. The sub-menu is not present.

chetnieter: So I found another potential solution.
```suggestion
  d->UI.menubar->setNativeMenuBar(false);
```
This moves most of the menu to the TeleSculptor window but it does allow the Options sub-menu to display.
Screen Shot 2021-09-13 at 12 35 00 PM

chetnieter: From the QT [documentation](https://doc.qt.io/qt-5/qmenubar.html#nativeMenuBar-prop).

mleotta: Yeah, I've seen that option, but it makes TeleSculptor look like a very non-standard Mac application.  I could be convinced otherwise, but I think I prefer the "0ptions" solution with zero.

chetnieter (approved):

PR #531, opened 2021-09-15
read camera resection config and use it

borovik135:
use camera resection configuration file for more flexible camera calibration capabilities


gui/MainWindow.cxx
```diff
@@ -1351,9 +1351,9 @@ void MainWindowPrivate::setActiveTool(AbstractTool* tool)
 
   auto const enableTools = !tool;
   auto const enableCancel = tool && tool->isCancelable();
-  foreach (auto const& tool, this->tools)
+  foreach (auto const& t, this->tools)
```
mleotta, line 1354: I don't think replacing `tool` with `t` is an improvement to readability.  I would drop this change.

gui/MainWindow.cxx
```diff
@@ -3572,7 +3572,11 @@ void MainWindow::computeCamera()
   auto landmarks = d->groundControlPointsHelper->registrationLandmarks();
 
   // Set up algorithm
-  auto config = kv::config_block::empty_config();
+  auto config = readConfig("resection_camera.conf");
```
mleotta, line 3575: for consistency, we should call this file `gui_resection_camera.conf`.  That naming convention may change in the future, but for now, all config files that are read directly from the TeleSculptor GUI via hard-coded file names in the GUI have a "gui_" prefix.

gui/MainWindow.cxx
```diff
@@ -3572,7 +3572,11 @@ void MainWindow::computeCamera()
...
+  if (!config)
+  {
+    config = kv::config_block::empty_config();
+  }
   config->set_value("algorithm:type", "ocv");
```
mleotta, line 3580: I suppose this is more of an issue with the existing code, but it seems odd to call this algorithm "algorithm".  it would be better to call it "resection" or something more informative.  That algorithm name appears both in code and config files.  If we named every algorithm "algorithm" there would be name clashes and we wouldn't know what each algorithm was for if they get merged into a common config file.

mleotta (changes requested):

gui/MainWindow.cxx
borovik135: ML, thanks for the comments; changes commited: add gui_ prefix to resection_camera.conf for consistency, use resection name for algorithm

gui/MainWindow.cxx
borovik135: name tool shadows the function parameter, which causes a warning, hence the change, not a big one, we can drop it here to address later

gui/MainWindow.cxx
borovik135: added gui_ prefix

gui/MainWindow.cxx
borovik135: now calling it "resection" instead of "algorithm"

gui/MainWindow.cxx
mleotta: ah, I see.  I missed that.  In that case it's fine to have here (or could just as easily be a separate PR,  we are due for a PR that just fixes a bunch of warnings)

mleotta (approved):
Looks good to me.  I targeted this to release

PR #535, opened 2021-09-20
Fix docker build

mleotta:


mleotta:
@chetnieter please review ASAP if you can.  This is the last change I need before I can make the TeleSculptor release.  

The dockerfile was out of date and wasn't building the latest code.  I had to update to a newer CMake, install a bunch of xcb packages required to build Qt 5.12.  Install sphinx to build the docs (installing the rtd theme did not pull in sphinx automatically).  I also updated the README to include the same package updates.

packaging/Dockerfile
```diff
@@ -15,13 +15,22 @@ ARG branch=release
 
 # Install required system packages to build TeleSculptor
 # Everything else is built from Fletch
-RUN apt-get update && apt-get install -y --no-install-recommends \
-build-essential file git cmake \
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive \
```
mleotta, line 18: Note: I had to add `DEBIAN_FRONTEND=noninteractive` here because sphinx requires `tzdata` which requires an interactive configuration of your timezone which hangs the docker build if you don't make it `noninteractive`.

chetnieter (approved):
Changes look fine to me. I am currently building the Dockerfile on my linux box. I will let you know that results of that when it is done.

PR #536, opened 2021-09-24
Add Flatpak and desktop and metainfo.xml file

kevinsmia1939:
These files are instructions to build TeleSculptor as Flatpak.
https://github.com/Kitware/TeleSculptor/issues/425


packaging/Flatpak/org.telesculptor.TeleSculptor.metainfo.xml
```diff
@@ -0,0 +1,39 @@
...
+            if available, from GPS and IMU sensors. However, the software can also work
+            with non-geospatial data and with collections of images instead of metadata.
+            TeleSculptor uses Structure-from-Motion techniques to estimate camera
+            parameters as well as a sparse set of 3D landmarks.
+            It uses Multiview Stereo techniques to estimate dense depth maps on key
+            frame and then fuses those depth maps into a consistent surface mesh which
+            can be colored from the source imagery.
+        </p>
+    </description>
+    <url type="homepage">https://github.com/Kitware/TeleSculptor</url>
```
mleotta, line 23: It might be better for this to be `https://telesculptor.org/`, which is the new homepage we've launched recently.  It links back to the Github page.

packaging/Flatpak/org.telesculptor.TeleSculptor.desktop
```diff
@@ -0,0 +1,11 @@
+[Desktop Entry]
+Version=1.0
+Name=TeleSculptor
+GenericName=Aerial Photogrammetry
+Comment=Photogrammetry designed with a focus on aerial video
+Exec=TeleSculptor %U
+Terminal=false
+Type=Application
+Icon=org.telesculptor.TeleSculptor
+Categories=Qt;Science;DataVisualization;
```
mleotta, line 10: We might consider adding `ImageProcessing`, `Geoscience`, and `3DGraphics`, maybe also `AudioVideo;Video`.  There are other categories that might fit too.

packaging/Flatpak/org.telesculptor.TeleSculptor.yaml
```diff
@@ -0,0 +1,88 @@
+app-id: org.telesculptor.TeleSculptor
```
mleotta, line 1: I think this is okay.  Another option might be `com.kitware.TeleSculptor`, but I'm not sure what rules are used to determine what this name should be.

mleotta (changes requested):
This looks reasonable to me, but I don't know much about Flatpak.  Please respond to the inline comments.  They are mostly suggestions.

packaging/Flatpak/org.telesculptor.TeleSculptor.yaml
kevinsmia1939: Since the new homepage is telesculptor.org, would not the org.telesculptor.TeleSculptor be better?

mleotta: Probably.  I'm not suggesting that this needs to be changed, I'm just trying to figure out how these ids are chosen and what the conventions are.  You said you've done this for other Kitware open source tools right?  Do those follow the same conventions?

packaging/Flatpak/org.telesculptor.TeleSculptor.desktop
kevinsmia1939: Does TeleSculptor also have audio, so it can be include in AudioVideo category?

packaging/Flatpak/org.telesculptor.TeleSculptor.metainfo.xml
kevinsmia1939: I will change it.

packaging/Flatpak/org.telesculptor.TeleSculptor.desktop
mleotta: Yeah, no audio, but it sounds like to include `Video` you also need `AudioVideo` as a super category.

packaging/Flatpak/org.telesculptor.TeleSculptor.yaml
mleotta: `org.telesculptor.TeleSculptor` seems correct then, especially if ParaView is `org.paraview.ParaView` since that is a highly analogous example.

barthalion:
@mleotta Sorry for hijacking this PR, but would you or anyone else from Kitware have write access to the TeleSculptor repo on Flathub? Kevin is doing a fantastic job maintaining apps, and so I imagine you wouldn't need to do anything specific, but one of Flathub principles is upstream involvement.

mleotta:
@barthalion I'm happy to be involved, but have not worked with Flathub yet myself.  @kevinsmia1939 has been pushing for Flatpak support and I'm happy to let him continue to do so, but if you need me to have access somewhere, just let me know what I need to do.

PR #542, opened 2022-03-16
Updated git ssh clone to comply with new standards

willdunklin:
Bugfix for errors cloning fletch. Github released these changes https://github.blog/2021-09-01-improving-git-protocol-security-github/ which finalized a breaking change on March 15, 2022.


mleotta (changes requested):
Good catch.  But lets use https instead.

mleotta (approved):

PR #543, opened 2022-03-17
Switch from write_pdal to pointcloud_io interface

willdunklin:
Replaced direct dependence of TeleSculptor code on PDAL with KWIVER vital `pointcloud_io` algorithm.


willdunklin:
@mleotta See my comment above. We actually want to wait for 1542 because it fixes a bug with file output.

mleotta:
Got it.  I missed that.  Thanks!

PR #548, opened 2024-03-27
Port TeleSculptor to Kwiver V2

aronhelser:
Main changes are set/get_nested_algo_config, create_algorithm and some API changes to mesh_coloration

Sprokit has not been enabled in kwiver V2, so the single Sprokit dependent class is disabled.

Cuda has not been tested.

Superbuild configs not yet updated - external fletch and kwiver v2 were used so far, with updated default versions for dependencies from kwiver v2


gui/WorldView.cxx
```diff
@@ -1811,10 +1812,7 @@ void WorldView::saveFusedMeshFrameColors(const QString &path, bool occlusion)
   coloration->set_frame_sampling(d->volumeOptions->getFrameSampling());
   double occlusionThreshold = d->volumeOptions->getOcclusionThreshold();
   coloration->set_occlusion_threshold(occlusionThreshold);
-  coloration->set_remove_occluded(occlusion);
-  vtkSmartPointer<vtkPolyData> meshFrameColors = vtkSmartPointer<vtkPolyData>::New();
```
mleotta, line 1815: Why are these lines removed?

aronhelser: `coloration->set_output()` was replaced with `get_output()`, and as far as I can tell, the only purpose of the `set_output()` was to provide a polydata for the results, which is now handled internally. I haven't examined the changed code in kwiver, though - the api changes came from https://github.com/Kitware/kwiver/pull/1534

aronhelser: yep, confirmed that creating the output polydata is handled internally using a ShallowCopy of the input, or other vtk pipeline result. https://github.com/Kitware/kwiver/pull/1534/files#diff-a4e99234c2dc9ae926db06dca4333e8dd8c49958ad50879ca8be3f45d1311794R215

aronhelser:
Updated with latest commit from "main" branch of kwiver. Once Chet confirms a successful build, should be GTG.

chetnieter:
I was able to build and launch TeleScuptor from the build directory but I get an error when I try and load a video.

```
WARN vital.algorithm: 
Implementation 'ffmpeg' for algorithm type video_input:type could not be found.
Make sure KWIVER_PLUGIN_PATH is set correctly.   Available implementations are:
      filter
      split
      image_list
      splice
      buffered_metadata_filter
      metadata_filter
      pos
```

This is after running `source setup_TeleSculptor.sh`. It appears to find the core implementations but not the ffmpeg ones. I checked and the ffmpeg arrow is being built.

chetnieter:
I also had issues with the installed version (installed with the `TeleSculptor-1.2.0-Linux-x86_64.sh` script). When I tried to run it I get this error.
```
./TeleSculptor: error while loading shared libraries: libQt5Xml.so.5: cannot open shared object file: No such file or directory
```

aronhelser:
I tried loading a video, and I saw when I started TeleSculptor that it couldn't find any of the plugin .so files. I uncommented the `export LD_LIBRARY_PATH` line with the fletch lib dir in `build/external/kwiver-build/setup_KWIVER.sh` and that was fixed. This indicates to me that the rpaths aren't fully working. I also encountered a python error about being unable to import `kwiver.vital.tests`, and that went away when I removed `export KWIVER_PYTHON_PLUGIN_PATH="kwiver.vital.tests.alg"` from the setup .sh

aronhelser:
Current state for me: superbuild shas updated for fletch and kwiver, and Telesculptor builds and runs for me from the build directory.

If I install the `TeleSculptor-1.2.0-Linux-x86_64.sh` script like Chet did, `./TeleSculptor` runs, but when I try to load a video it is unable to find any plugin implementations for loading the video. So the plugin path is not set correctly in this installer, seems like?

Work on this is currently paused in favor of work on burnout web.

aronhelser:
Tried again after the fletch path fix, and it seems like on video import the plugin error is fixed, but I get a different error: "ModuleNotFoundError: No module named 'kwiver'"

This seems like a Python error, and it is immediately after the console prints 
DEBUG - Initializing python interpreter
DEBUG - Python interpreter initialized

so it seems like the python path has a problem.

